PR #16, opened 2016-07-07
fix php white/blacklists in docker env

rainisto:
As docker has 'internal' 172.17. address, it failed the checks against docker machines external ip, so I added allow HTTP_HOST if server address is docker one.

### DIFF
--- a/alpine.docker
+++ b/alpine.docker
@@ -37,7 +37,9 @@ RUN mkdir -p /etc/pihole/ && \
     chown dnsmasq:root /var/log/pihole.log && \
     sed -i "s/@INT@/eth0/" /etc/dnsmasq.d/01-pihole.conf && \
     sed -i 's|"cd /etc/.pihole/ && git describe --tags --abbrev=0"|"cat /etc/pi-hole_version.txt"|g' /var/www/html/admin/footer.php && \
-    sed -i 's|"cd /var/www/html/admin/ && git describe --tags --abbrev=0"|"cat /etc/AdminLTE_version.txt"|g' /var/www/html/admin/footer.php
+    sed -i 's|"cd /var/www/html/admin/ && git describe --tags --abbrev=0"|"cat /etc/AdminLTE_version.txt"|g' /var/www/html/admin/footer.php && \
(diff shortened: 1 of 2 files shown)


diginc:
Hi, I was actually working on the same problem last night.

I believe the problem is already fixed in this pull request, which I do not have in any of my docker images yet: https://github.com/pi-hole/AdminLTE/pull/112 (sorry I haven't had time to test & release one since 1.7.0).  

Basically CORS was required to pass prior to that fix, which didn't go so well for non-chrome users who used anything but the 3 hostnames they allowed, now it is optional and as long as you have a php session from the pihole web interface you'll pass authorization.

I also have a fork going for AdminLTE where I also added this CORS logic fix along with my logging to help with debugging people hitting WL/BL issues: https://github.com/pi-hole/AdminLTE/pull/108

I haven't gotten the docker env -> webserver php env pass through completely automated yet for the other fix I have in there yet though...but having in the $VIRTUAL_HOST code doesn't hurt anything technically so I may end up baking a container off my fork anyway.

rainisto:
I looked on those changes, but afaik those don't fix the problem. For example my docker machine has 192.168.1.7 lan IP (shows as HTTP_ORIGIN and HTTP_HOST), and docker image has 172.17.0.3  internal ip (SERVER_ADDR) so checks would still give failed cors when accessing http://192.168.1.7/admin  it fails as 192.168.1.7 != 172.17.0.3, thats why I added the default docker networking range in the check (172.17.0.0/16).

diginc:
You're right, since I use a custom hostname for accessing pihole rather than IP I was more worried about getting support for arbitrary hostnames with the $VIRTUAL_HOST env var...and IP could be used in there but we already have ServerIP env var so I should just check CORS against that in addition to $VIRTUAL_HOST.

I don't have a ton of time right now though so I'll try to get this tested and merged today/tonight.

Thanks for the contribution.

diginc:
I just plowed ahead and finshed out supporting user set arbitrary environment variables for the hostname with $VIRTUAL_HOST env.  If this is not set it falls back to $ServerIP which should address your NAT docker0 issue.  

I built diginc/pi-hole:alpine_dev and diginc/pi-hole:debian_dev for you to try out to see if it addresses your issue.  It is staged on the dev branch for now, I wanted to perform a little more testing.

diginc:
@rainisto has this been working OK for you?

PR #22, opened 2016-07-18
py.test now builds containers, travis py.test

diginc:
* Instead of using the docker_dev, py.test builds and runs containers
* Travis-CI hopefully can run these same tests on PR/Merges now


diginc:
@jacobsalmela @dschaper @PromoFaux - would some rough automated integration tests like this be useful for pi-hole too?  Not exactly this obviously since my Dockerfile+build replaces the interactive installer for pi-hole.

This PR just got my new pytest suite running on travis-ci, the PR before this was the first py.test addition, the full code is view able here:
https://github.com/diginc/docker-pi-hole/tree/dev/test

I plan on growing the coverage with time to do things like:
- check that the procs include dns & web services 
- services respond as expected when queried
- whitelist/blacklist CLI and or web (maybe tricky w/authorization) function

Let me know if you have thoughts or suggestions on testing.  I'd be interested in collaborating on a python+docker test suite that can be applied to both my docker and your repos, but with different setup steps, e.i. your automated install instead of my docker build.

PromoFaux:
Hey man, test automation is something we've been thinking about, but we've been struggling to come up with a way to do that with bash. Plus, speaking for myself at least, not having any idea about testable code/CI/unit testing/whatever doesn't help... :)

Kind of putting off investing in travis if we don't really know what to do with it... but then.. if you're in a position to advise/help us out with setting something like that up..... What are your thoughts?

diginc:
It is possible to tack on tests to your existing project, for maximum value and early an proof of concept I'd recommend your tests start with a large scale end to end integration test (docker debian:jessie running basic-install.sh -> web/dns service running) and then you dive deep into smaller specialty tests that handle the various permutations possible within your core and helper scripts/functions.  

I would recommend doing this if the bash scripts will continue living for a while longer, otherwise you should probably just get lots of smaller unit tests in your python rewrite.  [pytest-flask maybe?](https://pypi.python.org/pypi/pytest-flask)

With the addition of an easy to use test framework, the tests should grow naturally (especially if your pull request template state passing/new tests are a requirement) as you run into bugs: e.g. write a test to catch the bug, fix said bug, then you have built in regression prevention going forward.

My docker image build gets rid of all user interactivity and lots of corner/edge case handling done by pi-hole's `basic-install.sh`.  I just give it one pathway to go in my docker image (well two, debian or alpine linux).  So you'd have to kind of do the same by having the test framework replace whiptail prompts with hard coded selections on a per-test bases (most likely with some test global defaults always applied if overrides aren't specified).

I'm interested in helping but may not have time to significantly support any testing efforts immediately but I'm available for questions/reviews (I could come to that slack I keep hearing about if it's not private).  

My current dev branch tests/ folder if fair game to any pi-hole members wanting a boilerplate docker + pytest framework.  Not a lot of code was required for pytest+testinfra to run a full integration test for my docker build + docker running.  ~1/3 of the code I committed is boilerplate example code form testinfra's docker example here: https://github.com/philpep/testinfra/blob/master/doc/source/examples.rst#test-docker-images and the rest was enhancement to conftest.py and the tests them selves.

PR #82, opened 2017-01-08
Ensure DNS is available early

mattkenn4545:
This change brings up dnsmasq early without relying on gravity update.  If dnsmasq is NOT started before gravity updates then it could be as long as the number of blocklist sources multiplied by 2m(timeout). Ie any of the blocklists being down could greatly effect the availability of the pi.hole service. 

This can be especially problematic if there are health checks running against the container... for example rancher/k8s etc... may reschedule the container after failing health checks :(

### DIFF
--- a/bash_functions.sh
+++ b/bash_functions.sh
@@ -153,17 +153,20 @@ test_framework_stubbing() {
 }
 
 main() {
+    echo -n '::: Starting up DNS and Webserver ...'
+    pihole restartdns # Just get DNS up. The webserver is down!!!
+
     IMAGE="$1"
-    case $IMAGE in
+    case $IMAGE in # Setup webserver
         "alpine")
-             gravity.sh # dnsmasq start included
-             php-fpm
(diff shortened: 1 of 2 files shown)


mattkenn4545:
I think I fixed the tests however Travis runs against the master branch...

diginc:
Thanks, this makes sense.  The tests appear to be running against your commit (travis says "Commit 622c778") - master is just the destination branch.

Could you change the destination merge branch to development or v2.11.1 ?  Any merge to master would auto re-build the old pi-hole version containers and I don't want to give them an impression that something new is out when it is just a patch to old containers.

I've been busy working on the update to the new version this week so v2.11.1 is still a WIP but that is the next thing I plan on merging to master.

mattkenn4545:
I don't think the tests are running against my commit ... the output from the failing travis job has master test names etc...

diginc:
see #84 for when this gets merged to master

PR #83, opened 2017-01-08
Use ip route to determine IP

mattkenn4545:


docker_dev.sh
```diff
@@ -3,7 +3,7 @@ docker build -f alpine.docker -t diginc/pi-hole:alpine .
 docker tag diginc/pi-hole:alpine diginc/pi-hole:latest
 docker build -f debian.docker -t diginc/pi-hole:debian .
 
-IP=$(ip addr show eth0 | grep "inet\b" | awk '{print $2}' | cut -d/ -f1)
+IP=$(ip route get 8.8.8.8 | awk '{ print $NF; exit }')
```
mat1th: Why don't you use the DNS1 environment variable?

mattkenn4545: Because DNS1 is only relevant inside the running container.  These scripts run on the host.

mat1th: Okay, Thanks for the explanation.

soredake: `ip route` gives ip adress of the tun0 interface, not the eth0
```
❯ ip route get 8.8.8.8 | awk '{ print $NF; exit }'                  
10.99.4.16
```

mattkenn4545: Where are you running the command?

soredake: Host machine, gentoo.

diginc: Knowing about the openVPN corner case I'm tempted to deny the change.  I know we could add more logic to say use this when the interface != tun0 but keep in mind I duplicate this command in the README (line 20) too, so I want it to be as short and concise as possible. 

Right now users who don't have eth0 probably know they don't have eth0 and just change NIC to en0/wlan0 or whatever.  I don't get a lot of feedback on the startup steps not working so it is partially a case of: if it isn't broke don't fix it too.

mattkenn4545: I see. This change simply finds what interface traffic to the outside world will egress from. Any network setup that changes this, like a vpn tunnel that changes the default route, will throw it. 

However, I think overall it makes it easier and more consistent for the general use cases of these scripts.

mattkenn4545: I'd posit that having a vpn tunnel is the exceptional configuration. 

Also, on Ubuntu 16.04+ network interfaces are 'consistent' ie the are NOT eth0.

diginc:
Out of curiosity everyone, especially @soredake :

Do you actually use the README or ./run_docker.sh command?  Or do you have a custom docker run / docker-compose setup?

I like the change @mattkenn4545 - I'm just short on data to know if this is more or less effective than what is out there currently so I'll probably just have to go with my gut.

soredake:
@diginc i use custo, docker.sh script to setup the containers.

mattkenn4545:
For me it's docker_dev. I don't use docker_run. 

It could be that changing the interface var is so easy that no one has complained. However that is a change many people have to make to get thing working. This change removes that step. 

What originally got me looking at this was that I'm on 16.04 so my nic is weirdly named. I also swapped between two boxes (with different nic names) and it was a pain having to manually change that file to dev efficiency.

docker_dev.sh
diginc: Good point, I didn't think about the deprecation of eth0 naming.  Going forward it is probably going to help more than hurt.  We should just maybe make it override able with a warning for VPN users

```
IP_LOOKUP=$(ip route get 8.8.8.8 | awk '{ print $NF; exit }') # May not work for VPN / tun0
IP=${IP:-$IP_LOOKUP} # if IP variable is set before running this script, it's value is used
```

mattkenn4545: Ya, I think this looks good.

diginc:
Thanks for the good discussion / contribution.  Merging, will be out with #84 's release/merge

README.md
```diff
@@ -16,12 +16,12 @@ One crucial thing to know before starting is the docker-pi-hole container needs
 
 ```
 IMAGE='diginc/pi-hole'
-NIC='eth0'
-IP=$(ip addr show $NIC | grep "inet\b" | awk '{print $2}' | cut -d/ -f1)
+IP_LOOKUP="$(ip route get 8.8.8.8 | awk '{ print $NF; exit }')"  # May not work for VPN / tun0
```
igalic: this is pretty cool, but it doesn't work on modern versions of iproute2 the last parameters on the line are `uid <uid number>`

diginc: Modern as in 4.X?
Could you provide a sample output from `ip route get 8.8.8.8`?

igalic:
```
8.8.8.8 via 192.168.0.1 dev wlp4s0 src 192.168.0.227 uid 1000 
    cache 
```

my iproute2 version is 4.15.0-2ubuntu1

diginc: Thank you, I'll open a new issue to document this bug

Issue #94, opened 2017-01-21
Error can't read /etc/pihole/setupVars.conf when mounting etc to volume


wtreur:
After the 2.11.1 update with #84 I tried to run the container as follows:

`docker run -v $(pwd)/etc/:/etc/pihole/ -e ServerIP=127.0.0.1 diginc/pi-hole:debian`

Since it is the first time, `$(pwd)/etc/` is empty and I get the following output

```
+ export IMAGE
+ export ServerIP
+ export ServerIPv6
+ export PYTEST
+ export PHP_ENV_CONFIG
+ export PHP_ERROR_LOG
+ export HOSTNAME
+ export WEBLOGDIR
+ export DNS1
+ export DNS2
+ export IPv6
+ . /bash_functions.sh
++ . /opt/pihole/webpage.sh
+++ setupVars=/etc/pihole/setupVars.conf
+++ dnsmasqconfig=/etc/dnsmasq.d/01-pihole.conf
+++ dhcpconfig=/etc/dnsmasq.d/02-pihole-dhcp.conf
+ echo ' ::: Starting docker specific setup for docker diginc/pi-hole'
+ validate_env
+ '[' -z 127.0.0.1 ']'
+ change_setting IPV4_ADDRESS 127.0.0.1
+ delete_setting IPV4_ADDRESS
+ sed -i /IPV4_ADDRESS/d /etc/pihole/setupVars.conf
 ::: Starting docker specific setup for docker diginc/pi-hole
sed: can't read /etc/pihole/setupVars.conf: No such file or directory
```

Workaround is to `touch etc/setupVars.conf` prior to running the container for the first time.

This is probably due to the changed made for the pi-hole password support. See: https://github.com/diginc/docker-pi-hole/pull/84#issuecomment-272341679

Same is true for `diginc/pi-hole:alpine`

diginc:
New version is working properly with empty volumes.  confirmed the admin interface can modify setupVars with the root permissions it is created with too.

PR #95, opened 2017-01-21
Ensured setupVars.conf is always present when changing settings from start.sh

wtreur:
Should solve #94


diginc:
Thanks, good catch!

PR #111, opened 2017-03-04
v2.13.1

diginc:
bump versions, fix shell check failures.


dschaper:
Sorry about that pointer release, needed to get that resolved quick and dirty as it was blocking new installs.

PR #138, opened 2017-05-24
Documentation updates

middleagedman:
Feel free to reject if needed.  Just saw somethings as I re-read instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,12 @@ The webserver and DNS service inside the container can be customized if necessar
 
 Similarly for the webserver you can customize configs in /etc/nginx (*:alpine* tag) and /etc/lighttpd (*:debian* tag).
 
+### Systemd init script
+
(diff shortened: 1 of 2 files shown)


diginc:
what is the benefit of systemd managing restarting the service over docker stop/start and --restart=always autobooting it?

diginc:
I'm not opposed to adding alternative ways to do the same thing if people want to do that.  I've been trying to slim down the readme with my recent changes rather than expand it though.

I've been thinking of setting up a github pages potentially to store multiple pages of documentation to reduce the readme clutter some.

middleagedman:
Well, normally you'd want to manage everything with systemd, since we are treating docker containers as a "service".  That way you can have dependencies.. I.e. don't start up the docker container before the dockerd is running, log additional items, etc.  But yea.. you can just put the service file in an etc folder or something.

diginc:
Going to merge, I may put some notes that this is functionality the same as --restart=always in later.

PR #140, opened 2017-05-29
Dev fixes for ARM & Alpine

diginc:


diginc:
@middleagedman I updated your systemd block to explain the equivalency to docker stop/start and --restart=always.  I linked to the script with master branch being the target so the link doesn't work until this PR is merged.

PR #142, opened 2017-06-01
use dnsmasq-dnssec

markschultz:
allows to enable dnssec in pi-hole settings

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -46,6 +46,7 @@ if [[ "$IMAGE" == 'debian' ]] ; then
 	apt-get install -y --force-yes netcat-openbsd
 elif [[ "$IMAGE" == 'alpine' ]] ; then
     apk add \
+        dnsmasq \
         dnsmasq-dnssec \
         nginx \
         ca-certificates \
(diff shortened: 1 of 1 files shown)


diginc:
How does this impact the users of the container?  [The tests appear to be failing](https://travis-ci.org/diginc/docker-pi-hole/builds/238242736?utm_source=github_status&utm_medium=notification) because this package change doesn't have the dnsmasq user installed like the standard dnsmasq package.

markschultz:
well pi-hole has a setting "Enable DNSSEC" currently if you enable that dnsmasq will continuously crash as it ws compiled without dnssec support. apologies, i couldn't actually tell what the error was. i've manually installed the `dnsmasq-dnssec` package on my local container and i can successfully enable that setting now.

diginc:
Gotcha, I get it now.  Debian probably already has the dnssec package since it uses the official installer so there probably isn't a change needed for those versions of the image.

If you didn't uninstall dnsmasq from your container then it sounds like dnsmasq-dnssec is an extra package and not a replacement for dnsmasq.  You should leave dnsmasq in the apk add command and just add the dnssec package after it.

I accidently deleted the dev branch while cleaning up merged branches.  I re-created it and set it as the target for merging.

diginc:
looks like I was wrong.  dnsmasq-dnssec and dnsmasq are just different conflicting packages on alpine apparently.  the latter comes with the dnsmasq user.

shaf:
Catered for in PR #175 

Cheers

Issue #154, opened 2017-07-26
Possible to lighttpd.conf server.bind = $ServerIP ?


FrozenDragoon:
Container created using the following string:
```
docker run -d \
--name pihole \
-h pihole \
--restart always \
-e INTERFACE=enp2s0f0:0 \
-e DNSMASQ_LISTENING=enp2s0f0:0 \
--net=host \
--cap-add=NET_ADMIN \
-v /srv/docker/dnsmasq:/etc/dnsmasq.d \
-v /srv/docker/pihole:/etc/pihole \
-v /var/log:/var/log \
-e TZ=America/Chicago \
-e WEBPASSWORD="snip" \
-e ServerIP=192.168.2.50 \
-e IPv6=False \
diginc/pi-hole:debian
```
----

My host has two interfaces, one that I am using for PiHole.  I also run an internal web server on this host, and for my application port 80 is required.

When running Pihole via normal install this is not a problem, as I just bound all pihole functions to the second interface. 

The container's Lighttpd automatically attempts to bind "*:80".  Is it possible to add the [server.bind](https://redmine.lighttpd.net/projects/1/wiki/Server_bindDetails) config line using the ServerIP environment variable? This will allow lighttpd to not attempt to overtake port 80 on every interface.

I have made the change manually, and can verify it works.  

Other than this, the container is working flawlessly so far.  Thanks for your work!

FrozenDragoon:
```sed -i 's/server.port                 = 80/server.bind\t\t    = ${ServerIP}\nserver.port\t\t    = 80/' test.txt```
That works in pretty basic testing.  Maybe it could be added to ```docker-pi-hole/debian/start.sh```? 

This shouldn't cause a conflict due to ServerIP being a required variable. 

I'll see if I can put together a pull request, though I am very new to using git collaboratively.

diginc:
Give the dev tags a shot to confirm it built in the cloud correctly (I forget that sometimes and just run my local docker builds).  Thanks for the report & fix.

FrozenDragoon:
Pulled down the dev container and it all appears to be running smoothly, thanks for the quick response

PR #155, opened 2017-07-26
Bind lighttpd to only one IP when using "--net=host"

FrozenDragoon:
See #154. 

This only applies to the Debian image.  I imagine it would be possible to do something similar for nginx on the Alpine image, though.

### DIFF
--- a/bash_functions.sh
+++ b/bash_functions.sh
@@ -119,16 +119,16 @@ setup_dnsmasq_hostnames() {
 }
 
 setup_lighttpd_bind() {
-if [[ "$IMAGE" == 'debian' ]] ; then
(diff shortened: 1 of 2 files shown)


bash_functions.sh
```diff
@@ -118,6 +118,19 @@ setup_dnsmasq_hostnames() {
     fi
 }
 
+setup_lighttpd_bind() {
+if [[ "$IMAGE" == 'debian' ]] ; then
+	# if using '--net=host' only bing lighttpd on $ServerIP
+	HOSTNET='grep "docker" /proc/net/dev/' #docker (docker0 by default) should only be present on the host system
```
diginc: You have an extra slash on the end of `/proc/net/dev`

PR #160, opened 2017-09-03
Fix passing in DNS1 and DNS2 env vars to setup_dnsmasq

zpartal:
Noticed that DNS1 and DNS2 env variables were not being picked up by pihole. Discovered this while trying to debug why my custom DNS set in the web admin was being reset every time the container restarted.


diginc:
This was merged to master but I also pushed it to dev branch to get a dev image built.  master based docker image tags are a manual trigger to build so the official tags won't be out until after I do a quick vet of the official docker cloud built `dev` images.

Issue #156, opened 2017-07-26
'Website Blocked' page reports incorrect domain


Tatsh:
Instead of using the domain of the site blocked, it says *dockerpi.hole*.

I am running like so in a shell script:

```bash
docker run \
    -p 53:53/tcp \
    -p 53:53/udp \
    -p 8100:80 \
    --cap-add=NET_ADMIN \
    -e ServerIP="$IP" \
    -e WEBPASSWORD='somepass' \
    -e DNS1=192.168.1.1 \
    --restart=always \
    --name pihole \
    -v '/home/tatsh/dev/pihole-config/:/etc/pihole/' \
    -v '/home/tatsh/dev/pihole-dnsmasq.d/:/etc/dnsmasq.d' \
    -v /sys/fs/cgroup:/sys/fs/cgroup:ro \
    -d \
    diginc/pi-hole
```

My configuration is nearly default except for a `blacklist.txt` file.

Tatsh:
In the code `/var/www/html/pihole/index.php`, the line is: `$serverName = escapeshellcmd($_SERVER['SERVER_NAME']);` so it seems like it is taking this information from nginx, and nginx is passing the wrong value to PHP-FPM. Maybe this is because of the explicit `server_name` value set to *dockerpi.hole*. Still experimenting.

1activegeek:
I'm running into this issue as well. I was hoping the VIRTUAL_HOST variable would set this properly, but it doesn't seem to be doing the trick. I'm hoping someone can help here, because I want to be able to use this as an easy way for the significant other and I to easily add sites when needed. Right now, it's a complicated explanation of how to login to the pihole and complete the task. Hitting a block page, putting in the password, and being on the way would make it much easier.

1activegeek:
So it would still be good to get a resolution to this, but as an alternate work around - use the debian tagged version of the container. It seems that one is running as intended, and is properly grabbing the URL and allowing for use of the whitelist function.

Lyrex:
I played a bit with the variables and found the following working for me (using the default alpine linux container):  
Use `$serverName = escapeshellcmd($_SERVER['HTTP_HOST']);` instead of the default `$serverName = escapeshellcmd($_SERVER['SERVER_NAME']);`

This should work since most proxies (including jwilder in it's default config [[1]](https://github.com/jwilder/nginx-proxy#replacing-default-proxy-settings)) forward the `Host` header to the destination server.

1activegeek:
[...]

PR #161, opened 2017-09-05
Fix "wrong domain bug" on block page

Lyrex:
This fixes issue #156 which says that on the block page, at least on alpine linux display the wrong domain which makes easy whitelisting not possible.


diginc:
Re-built the failed alpine image tests - seems like it was a transient apk server issue.  Look good, merging to diginc/pi-hole:alpine_dev for testing

Issue #165, opened 2017-10-01
Alpine based image for ARM


klud1:
It'd be good if an alpine based image for ARM was available, the footprint will be really small compared to the current one, the alpine Dockerfile can be used as sample for it.

#166

PR #166, opened 2017-10-01
Alpine based images for ARM*

klud1:
Reduce the image footprint. If accepted, README stils needs to be updated in order to add the new tags.
Related issues:
#165

### DIFF
--- a/alpine-arm.docker
+++ b/alpine-arm.docker
@@ -1,0 +1,60 @@
+FROM multiarch/alpine:armhf-edge
+MAINTAINER adam@diginc.us <adam@diginc.us>
+
+ENV IMAGE alpine
+ENV PATH /opt/pihole:${PATH}
+
+COPY install.sh /usr/local/bin/docker-install.sh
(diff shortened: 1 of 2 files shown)


diginc:
Cool thanks for the new versions.  I'm going to change this PR to merge to dev branch so we can test building and running tags.  

From my testing so far these multiarch images appear to be the same qemu magic as my debian resin.io base image but much faster executing...nice find, not sure why I never gave these ones a shot but I might change the debian arm over to it now.

klud1:
@diginc Hi, ya. I just found an issue on them, the pihole-FTL binary used was built for amd64, so it causes a exec format error, gonna check the scripts and modify/duplicate them for them to use the proper binary

klud1:
It pulls the amd64 bin only, time to do some logic there

https://github.com/diginc/docker-pi-hole/blob/1baf6a5258da2f46de23c46e82c89bc6ce4bf04a/install.sh#L29

Issue #171, opened 2017-10-12
no DNSSEC compiled in alpine


shaf:
Hi,

DNSSEC fails due to the following ....

docker exec -it pihole-shaf bash
bash-4.3# dnsmasq -v
Dnsmasq version 2.78  Copyright (c) 2000-2017 Simon Kelley
Compile time options: IPv6 GNU-getopt no-DBus no-i18n no-IDN DHCP DHCPv6 no-Lua TFTP no-conntrack ipset auth no-DNSSEC loop-detect inotify

Please use the following package instead

bash-4.3# apk search dnsmasq
dnsmasq-2.78-r0
dnsmasq-dnssec-2.78-r0

diginc:
Related: https://github.com/diginc/docker-pi-hole/pull/142

There seemed to be build issues when trying to install dnsmasq-dnssec apk.  I never looked any further than the travis-ci failures in that PR.

shaf:
Thanks mate, I'll take the PR

diginc:
I haven't confirmed DNSSec + alpine operation personally yet, if anyone else has let me know.

Automated python tests will hopefully be implemented at some point.

diginc:
I've confirmed that #175 didn't quite fix the issue.

When I check the 'enable DNSSec' checkbox in the web interface while watching the docker logs it restarts dnsmasq and immediately starts throwing 'DNSMasq not compiled with dnssec' errors.  apk show lists `dnsmasq` packaged rather than `dnsmasq-dnssec` too

PR #175, opened 2017-10-16
#171,#142 enable DNSSEC for alpine

shaf:
First I committed to master (sorry bad habit)

Second FAILED

Third time lucky https://travis-ci.org/shaf/docker-pi-hole/builds/288786322


diginc:
Clever, that'll do the trick.

PR #184, opened 2017-11-02
escape a pipe character which was breaking a table

phelma:


diginc:
Thanks, can't believe I missed that one.

PR #214, opened 2018-01-10
Add a healthcheck

oscarnyl:
I noticed that my container stayed alive despite DNS resolution no longer working at all. This should rectify that problem and restart the container if DNS resolution is available.

Note that I havent tested this with this Dockerfile, I have only added the equivalent to my own docker-compose file (and so long it appears to be working fine). www.google.com is just a common address I decided to test with, I'm completely fine with any other being the healthcheck name


diginc:
`pi.hole` maybe a better address to check.

diginc:
if we're going to add this we may want to change `--restart=always` to `--restart=unless-stopped` to prevent trashing between restarting always running and the health check stopping the container.

oscarnyl:
Good idea on resolving pi.hole.

dig returns zero for NXDOMAIN so even if someone were to remove their pi.hole alias this will still work out.

oscarnyl:
I read the documentation again for Dockerfile HEALTHCHECK, and it turns out that this is a better way to do the healthcheck. dig will not necessarily return 1 on bad health, but this will force a return 1 on non-zero exit which was the behaviour I was looking for.

PR #216, opened 2018-01-13
[Untested] Make Lighttpd also bind to localhost when using host network

DerFetzer:
When using `--net=host` in a setup without proxy blockingpage is broken.
This is due to Lighttpd only binding to `$ServerIP` and not to localhost.

I've tested the change locally.


diginc:
Automated tests passed, looks good.  I'll merge and pull it down from the dev tag to test in docker bridged mode.

Issue #217, opened 2018-01-13
(mod_fastcgi.c.1406) [ERROR]: spawning fcgi failed


xenithorb:
```
2018-01-13 13:52:17: (log.c.164) server started
2018-01-13 13:52:17: (mod_fastcgi.c.984) bind failed for: unix:/var/run/lighttpd/php.socket-0 No such file or directory
2018-01-13 13:52:17: (mod_fastcgi.c.1406) [ERROR]: spawning fcgi failed.
2018-01-13 13:52:17: (server.c.1022) Configuration of plugins failed. Going down.
```

Getting this in /var/log/lighthttpd/error.log 

`docker logs` shows s6 is restarting lighthttpd over and over again, but otherwise pi-hole setup was successful and DNSmasq is running and working. FTL is also running OK. 

```
REPOSITORY                 TAG                 IMAGE ID            CREATED             SIZE
docker.io/diginc/pi-hole   latest              6b77d5329fd4        8 days ago          330.5 MB

**Comment by diginc** [comment]
Are there any volumes pointed to /var/?  What's the container run configuration look like?

**Comment by vmtyler** [comment]
+1 on this issue. should we put a PR in for the temp fix?

**Comment by diginc** [comment]
I've been out sick for a week, trying to reproduce with docker_run.sh now.

**Comment by diginc** [comment]
I was unable to reproduce.  Can I see a docker run command example?  Also what OS & Docker version are those of you seeing this problem using?

I ran this exactly: https://raw.githubusercontent.com/diginc/docker-pi-hole/master/docker_run.sh

```
$ docker inspect pihole | grep -i image
        "Image": "sha256:6b77d5329fd484527ec719e52e3f2c85cecea0a06a652feeed8779f458d8abd0",
            "Image": "diginc/pi-hole:latest",
                "image": "diginc/pi-hole:debian_amd64",
```

/var/run/lighttpd permissions:

```
$ docker exec pihole ls -lat /var/run/lighttpd
total 8
drwxr-xr-x 2 www-data www-data 4096 Jan 25 15:17 .
drwxr-xr-x 8 root     root     4096 Jan 25 15:17 ..
srwxr-xr-x 1 www-data www-data    0 Jan 25 15:17 php.socket-0
```

**Comment by vmtyler** [comment]
```
docker inspect pihole | grep -i image
        "Image": "sha256:6b77d5329fd484527ec719e52e3f2c85cecea0a06a652feeed8779f458d8abd0",
[...]

PR #227, opened 2018-01-30
Use pihole checkout for development checkout

diginc:
And add FTL

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -52,9 +52,11 @@ installPihole | tee "${tmpLog}"
 mv "${tmpLog}" /
 
 if [[ $USE_DEVELOPMENT_BRANCHES == true ]] ; then
-    pihole checkout core development
-    pihole checkout web devel
-    pihole checkout ftl development
+    ln -s /bin/true /usr/local/bin/service
+    echo y | bash -x pihole checkout core development
+    echo y | bash -x pihole checkout web devel
+    echo y | bash -x pihole checkout ftl development
+    unlink /usr/local/bin/service
 else
     # Reset to our tags so version numbers get detected correctly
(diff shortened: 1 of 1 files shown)


HorayNarea:
FTL is not starting correctly since I updated to the newest dev-image (which contains this merged PR)
It just keeps spamming the log with "Starting pihole-FTL" and never gets started -> no data shown in the webinterface

HorayNarea:
seems like the directory for its socket is missing:
```
docker exec -it pihole /bin/bash
ls /var/run/pihole
ls: cannot access '/var/run/pihole': No such file or directory
```

diginc:
#227 Had a fix for that one @HorayNarea

HorayNarea:
Well my docker host is an Archlinux… :D

diginc:
Is it doing any sort of weird /var mounts like described in #217 ?  That is what fedora was doing and it was breaking the expected directory structure of /var.  You could do a docker inspect on your container to see if there's a volume mount for /var that you did not add.

```
docker pull diginc/pi-hole:debian_dev ; docker run -it -e ServerIP="1.2.3.4" -v /dev/null:/etc/pihole/adlists.default -e PYTEST="True" diginc/pi-hole:debian_dev bash
...
root@f0adfa825680:/# ls -lat /var/run/pihole
total 8
drwxr-xr-x  2 root root 4096 Jan 31 19:50 .
drwxr-xr-x 11 root root 4096 Jan 31 19:50 ..
srwxr-xr-x  1 root root    0 Jan 31 19:50 FTL.sock
```

HorayNarea:
Nope, just `/etc/pihole` and `/etc/dnsmasq.d`
bildschirmfoto vom 2018-01-31 21-05-25

PR #237, opened 2018-03-02
Develoment to Master

diginc:
* adding FTL_CMD for debug


bobsoppe:
Fix below should enable custom FTL arguments with a sane default
`export FTL_CMD="${FTL_CMD:-no-daemon}"`

PR #253, opened 2018-04-05
quick readability edits

tedder:
lots of spelling fixes, add a second header so it doesn't blend together


diginc:
Thanks again!

PR #254, opened 2018-04-05
copyedit, round two

tedder:
more copyedits.


diginc:
Thanks for fixing my sloppy typos :D

Issue #272, opened 2018-05-18
pi-hole admin page not showing statistics


mdclemen:
This is a... 

- [ ] Request for a new or modified feature
- [x ] Issue trying to run the docker image
- [ ] Issue trying to build / test / develop the docker image

## Description

I am able to successfully run the pi-hole docker image, but my admin page is not showing statistics. I am running the docker image on my laptop (linux mint 17.3/4.4.0-72-generic) and the output of docker logs pihole is:
### `[s6-init] making user provided files available at /var/run/s6/etc...exited 0.
[s6-init] ensuring user provided files have correct perms...exited 0.
[fix-attrs.d] applying ownership & permissions fixes...
[fix-attrs.d] 01-resolver-resolv: applying... 
[fix-attrs.d] 01-resolver-resolv: exited 0.
[fix-attrs.d] done.
[cont-init.d] executing container initialization scripts...
[cont-init.d] 20-start.sh: executing... 
ARCH=amd64
BASH=/bin/bash
BASHOPTS=cmdhist:complete_fullquote:extquote:force_fignore:hostcomplete:interactive_comments:progcomp:promptvars:sourcepath
BASH_ALIASES=()
BASH_ARGC=()
BASH_ARGV=()
BASH_CMDS=()
BASH_LINENO=([0]="0")
BASH_SOURCE=([0]="/var/run/s6/etc/cont-init.d/20-start.sh")
BASH_VERSINFO=([0]="4" [1]="4" [2]="12" [3]="1" [4]="release" [5]="x86_64-pc-linux-gnu")
BASH_VERSION='4.4.12(1)-release'
CWD=/
DIRSTACK=()
EUID=0
FTL_CMD=no-daemon
GROUPS=()
HOME=/root
HOSTNAME=6250e570f0fe
HOSTTYPE=x86_64
IFS=$' \t\n'
IPv6=True
MACHTYPE=x86_64-pc-linux-gnu
OPTERR=1
OPTIND=1
OSTYPE=linux-gnu
PATH=/opt/pihole:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
PHP_ENV_CONFIG=/etc/lighttpd/conf-enabled/15-fastcgi-php.conf
PHP_ERROR_LOG=/var/log/lighttpd/error.log
PIHOLE_INSTALL=/tmp/ph_install.sh
PPID=208
PS4='+ '
PWD=/
[...]

Issue #241, opened 2018-03-09
pihole-FTL: no process found


skalg:
Im using the latest image and i got pihole-FTL: no process found with a clean config/volume.

Here's the last lines of logs
```

pihole_1  | 
pihole_1  | pihole-FTL: no process found
pihole_1  |   [✓] DNS service is running
pihole_1  |   [✓] Pi-hole blocking is Enabled
pihole_1  | [cont-init.d] 20-start.sh: exited 0.
pihole_1  | [cont-init.d] done.
pihole_1  | [services.d] starting services
pihole_1  | Starting pihole-FTL
pihole_1  | Starting dnsmasq
pihole_1  | Starting crond
pihole_1  | [services.d] done.
pihole_1  | Starting lighttpd
pihole_1  | dnsmasq: started, version 2.76 cachesize 10000

```

And /admin/api.php?overTimeData10mins gives me FTLnotrunning | true

diginc:
Do you have a docker run command to reproduce this problem?

pascalgoty:
Here's my docker-compose, i've also tried with volumes for /etc/pihole and /etc/dnsmasq.d

version: "3"
services:
  pihole:
    image: diginc/pi-hole:latest
    ports:
      - 53:53/tcp
      - 53:53/udp
      - 80:80/tcp
    dns: 127.0.0.1
    volumes:
      - /etc/localtime:/etc/localtime:ro
    cap_add:
      - NET_ADMIN
    environment:
      ServerIP: 192.168.0.2
      TZ: America/Toronto
      DNS1: 209.222.18.222
      DNS2: 209.222.18.218
      WEBPASSWORD: 1234
    restart: always

nickbabcock:
Same issue. FTL not running so no graphs (pihole itself still works). Also using docker-compose. I was updating from `diginc/pi-hole:debian_v3.2.1`. Pihole was also pegged at 100% CPU usage for some reason.

```yaml
version: '3'
services:
  pihole:
    container_name: 'pihole'
    image: 'diginc/pi-hole:debian_v3.3'
    restart: 'unless-stopped'
    ports:
      - '53:53/tcp'
      - '53:53/udp'
      - '8182:80'

    # For me this env file needs ot have WEBPASSWORD, ServerIP, VIRTUAL_HOST
    env_file: pihole-variables.env
    environment:
      IPv6: 'False'
    volumes:
      - '/tank/containers/pihole/etc:/etc/pihole/'
[...]

PR #287, opened 2018-07-02
Fail container on error in start.sh

StefanSchoof:
## Description
Fail the container if a error in the start.sh occurs.

## Motivation and Context

If the container is started with work parameters or something is going wrong in the start.sh the container regardleas spins up. This can result into direct errors or errors longer down the way. 
For example 
`docker run --rm -v "$(pwd)/dnsmasq.d/:/etc/dnsmasq.d/" pi-hole-multiarch:debian_armhf` result into a running container and creates a `01-pihole.conf` with only `addn-hosts=/etc/pihole/gravity.list` as content. This results into missing log entries and no stats. See #272 and #241. Possible are other problems.

## How Has This Been Tested?
Run the image without an ServerIP and with an ServerIP environment variable. Without ServerIP the container now fails and with an ServerIP the container spins up.

## Types of changes

## Checklist:

### DIFF
--- a/s6/debian-root/etc/cont-init.d/20-start.sh
+++ b/s6/debian-root/etc/cont-init.d/20-start.sh
@@ -1,5 +1,6 @@
 #!/usr/bin/with-contenv bash
 set
+set -e
 
 bashCmd='bash -e'
 if [ "${PH_VERBOSE:-0}" -gt 0 ] ; then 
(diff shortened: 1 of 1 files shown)


diginc:
I Retried the travis-ci test build to see if it's flakiness or an actual error.  I ran the tests locally a couple days ago and they all came back green.

diginc:
Green now, `-e` can go on line 2 instead of a new line 3.

I had removed it from line 2 while debugging a test error and forgot to re-add.

s6/debian-root/etc/cont-init.d/20-start.sh
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/with-contenv bash
 set
```
diginc: `-e` can go on line 2 instead of a new line 3.

PR #289, opened 2018-07-05
Prerelease of FTLDNS image

diginc:
## Description
Pi-Hole FTLDNS pre release changes

## Motivation and Context
Prep work to confirm the image is ready for FTLDNS

## How Has This Been Tested?
pytests updated.  Running resulting image for personal use.

## Types of changes
New version!

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,4 +21,5 @@ script:
   - ./Dockerfile.py --arch=${ARCH} -v
   - docker images
   # run docker build & tests
-  - py.test -vv -n auto -k "${ARCH}" ./test/
+  # 2 parallel max b/c race condition with docker fixture (I think?)
+  - py.test -vv -n 2 -k "${ARCH}" ./test/
--- a/Dockerfile.py
+++ b/Dockerfile.py
@@ -27,7 +27,7 @@ THIS_DIR = os.path.dirname(os.path.abspath(__file__))
 base_vars = {
     'name': 'diginc/pi-hole',
     'maintainer' : 'adam@diginc.us',
-    's6_version' : 'v1.21.2.2',
+    's6_version' : 'v1.21.4.0',
 }
 
 os_base_vars = {
@@ -111,6 +111,7 @@ def build(docker_repo, os, arch, args):
     build_result = run_local(build_command) 
     if args['-v']:
         print build_result.stdout
+        print build_result.stderr
     if build_result.rc != 0:
         print "     ::: Building {} encountered an error".format(dockerfile)
         print build_result.stderr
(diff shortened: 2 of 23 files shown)


diginc:
Available test images:

For [x86](https://hub.docker.com/r/diginc/pi-hole/tags/)
  - `diginc/pi-hole:prerelease`
- For [ARM](https://hub.docker.com/r/diginc/pi-hole-multiarch/tags/) per https://github.com/diginc/docker-pi-hole/issues/295
  - `diginc/pi-hole-multiarch:debian_aarch64_prerelease`
  - `diginc/pi-hole-multiarch:debian_armhf_prerelease`

diginc:
Reminder: contingent on the next release not using block pages, but NULL DNS responses by default, I may remove the validation of "ServerIP" being set and exiting the container startup when not set since it will not be required anymore.

nightah:
Looks like some of the default behaviour like grabbing and updating the default adlists doesn't work in this pre-release, not sure if that's by design or an oversight.

Happy to run it for longer and provide other feedback if necessary.

diginc:
There are [no adlists by default in the new FTLDNS pi-hole installer](https://github.com/pi-hole/pi-hole/blob/FTLDNS/automated%20install/basic-install.sh#L992), there is a prompt for users to select the current 'recommended' ones they want to opt into.  I haven't made an equivalent way for users to opt-in to the lists of ads.  I'll add something for this and do more checking on the updating.

Once the lists are selected they seemed to update OK during container startup and the web interface, I'm just re-using my old 3.3.1 container volume data to seed the lists in all real world tests so far though.

I saw the pihole-FTL restart errors too, I am pushing a fix for those now.

nightah:
I had previously specified the dns server for the pi-hole container as `127.0.0.1` and that seemed to have issues when I re-used my old container volume. When I removed that setting things seem to be working again.

Few other issues I seem to encounter with the new version:

- Query Log is stuck on `Processing...` and does not show any recent queries.
- Tools -> Tail pihole.log results in the following error: `Failed to open log file. Check permissions!`
I suspect this is because it's running as UID/GID `root / dip` vs `root / root`

All the other features seem to more or less work appropriately with my limited testing.

nightah:
Also noticed that the both Query Types and Queries answered by pie charts on the Dashboard appear to have missing pieces.

You can see [here](https://i.imgur.com/f17sYqH.png) what I mean. Again could be unrelated but that doesn't seem to be an issue on the main release. Happy to lodge this upstream if you're experiencing the same problem.

diginc:
Moving this branch to match pi-hole's release branch name scheme (release/v4.0) and made a new pull request: #300

PR #290, opened 2018-07-06
allow no for DNS2

### LINKED ISSUE #281: Have only a single DNS Server
This is a... 

- [x] Request for a new or modified feature
- [ ] Issue trying to run the docker image
- [ ] Issue trying to build / test / develop the docker image

## Description
In the [pihole configuration guide for routers](https://discourse.pi-hole.net/t/how-do-i-configure-my-devices-to-use-pi-hole-as-their-dns-server/245) has the second option ("Advertise Pi-hole’s IP address via dnsmasq in the router (if supported)") the section:

> On Pi-hole, login to the web interface http://pi.hole > Settings > DNS and instead of choosing upstream servers like Google or OpenDNS, set the upstream to be the IP address of the router as the only upstream DNS server. Do not define any other DNS entries for Pi-hole:

I found no way to do this in the docker image. In the [Environment Variables](https://github.com/diginc/docker-pi-hole#environment-variables) section is mention how to unset DNS2 and in a quick look into the [bash_functions.sh](https://github.com/diginc/docker-pi-hole/blob/8e2889ba561121a3644d97481137033aa19bdd2f/bash_functions.sh#L41) I see with my bash knowlage no way to do this.

## Expected Behavior
Having a way to have not default the second dns to 8.8.4.4.

## Actual Behavior
If no value is used for the second dns, it is always 8.8.4.4.

## Possible Fix
Do not set second dns if it has the value "0.0.0.0" or "unset".

StefanSchoof:
## Description
Allows to have only a single DNS Server and not default to the second Google DNS.

## Motivation and Context
Resolves #281 

## How Has This Been Tested?
Run py.test on armhf (raspberry pi 3) with the new tests and run the container with DNS2 `no` and with DNS2 with the value `1.1.1.1`

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ There are other environment variables if you want to customize various things in
 | `-e TZ=<Timezone>`<br/> **Recommended** *Default: UTC* | Set your [timezone](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) to make sure logs rotate at local midnight instead of at UTC midnight.
 | `-e WEBPASSWORD=<Admin password>`<br/> **Recommended** *Default: random* | http://pi.hole/admin password. Run `docker logs pihole \| grep random` to find your random pass.
 | `-e DNS1=<IP>`<br/> *Optional* *Default: 8.8.8.8* | Primary upstream DNS provider, default is google DNS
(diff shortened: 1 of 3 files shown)


diginc:
Thanks for the contribution!

PR #302, opened 2018-08-04
Repo naming changes

### LINKED ISSUE #301: updated references to old image path
Reminder to update all docs and docker run and docker compose in the repo.  New image path intentionally not put here as it is not 'released' yet.

diginc:
resolves #301

There were also a bunch of remnants of Alpine VS Debian image, and Debian still being in the tag didn't make sense to port over with the repo/org change.

The multi-arch-images are not a seperately named image anymore either since we're not doing a docker inc. 'Automated Build' we do not have restrictions imposed by that anymore.  That's not to say this image's build isn't going to be automated, it will be.


diginc:
Tests are still failing because of debian being in the tags.  Having them as regression testing is important during refactoring so I'll address that in this PR too.

```
E       Failed: Unexpected exit code 125 for CommandResult(command='docker run -d -e ServerIP="1.2.3.4" -e DNS1="1.2.3.4" -e DNS2="2.2.3.4" --dns 127.0.0.1 -v /dev/null:/etc/pihole/adlists.default -e PYTEST="True" pi-hole:debian_amd64 tail -f /dev/null', exit_status=125, stdout=None, stderr="WARNING: Localhost DNS setting (--dns=127.0.0.1) may fail in containers.\nUnable to find image 'pi-hole:debian_amd64' locally\ndocker: Error response from daemon: pull access denied for pi-hole, repository does not exist or may require 'docker login'.\nSee 'docker run --help'.\n")
```

Issue #243, opened 2018-03-17
[ERROR] sudo: unable to send audit message on Synology


heseifert:
Hello,
i got this error on my docker logs.

`sudo: unable to send audit message: Operation not permitted`

If i do anything on the WebUI i got this error.

My Setup:
- Synology DSM 6.1.5 DS1515+
- Lastest docker-pi-hole Image

Can any help me?

diginc:
I would start with synology + that error message unless this is very specific to my container.

have you looked at these posts?  https://duckduckgo.com/?q=site%3Aforum.synology.com+Operation+not+permitted&t=ffab&ia=web

Do you use volumes and if you remove them and re-create the container does this error go away?

jcacavias:
I'm getting a similar error message each time I try to press the enable button.  Enable appears to work, but as soon as I refresh the page, the enable button is back.  Otherwise, pi-hole appears to be working.  Here are the errors that appear in the logs...

sudo: unable to send audit message: Unknown error -1
sudo: pam_open_session: System error | stdout
sudo: policy plugin failed session initialization

edudojevic:
Same here when I try `sudo pihole -r`:
```
root@diginc-pi-hole1:/# sudo pihole -r
sudo: unable to send audit message: Unknown error -1
sudo: pam_open_session: System error
sudo: policy plugin failed session initialization
```

Also having the same problems through the Web UI, where I cannot save anything. It seems like a permissions issue, but since I'm using the already built diginc pihole docker image I would expect this to work out of the box. Running this on Synology DS916+.

Any help is appreciated.

ShagoY:
Hi, i got this error only with --net=host (the only way to see "Top Clients ?")

kn0rki:
Hi folks,
same errors for me on a DS412+.

@ShagoY the --net=host works without NAT. You can add a new network and disable ip masquerading to enable the top clients.

ShagoY:
Hi @kn0rki, i just try in docker but no result, i think i need more information, have you a guide ?

Tearow:
Same issue.
When ever i try to add something in whitelist or add something in blacklist or any setting i receive in docker terminal.

Sudo: Unable to send audit message: Unknown error -1
Sudo: pam_open_session: System error
Sudo: policy plugin failed session initialization

did somebody solve it already. i saw already multiple issues about this but still without solution.

Using DS412+ with docker and Diginc/pi-hole:latest

[...]

PR #358, opened 2018-10-30
Synology sudo audit error fix, disable audit in pam modules

diginc:
## Description
Older kernels get some fatal sudo audit errors in docker images that utilize su/sudo, which is sprinkled throughout pi-hole.

## Motivation and Context
Fix it

## How Has This Been Tested?
Automated tests, lots of people testing in issue https://github.com/pi-hole/docker-pi-hole/issues/243

## Types of changes
Moved non pi-hole stuff to a new setup.sh script, split it into it's on RUN step because it is a REALLY long manual build of PAM modules and docker caching is going to be required to maintain sanity.

Increased the docker image size, but haven't had time to hunt down where that size is.  Still running the clean up of /var/lib/apt and also added a deletion of the manually re-built debs for pam.

## Checklist:
- [?] My change requires a change to the documentation.
- [?] I have updated the documentation accordingly.


dschaper:
Would removing `sudo` from the calls if the running user was root help?

diginc:
As I typed that description the same idea occurred to me, I think it's `su` as much as `sudo` after reviewing some of the original links about the issue:

- https://github.com/moby/moby/issues/5899 - noted this comment about it being [more than just su](https://github.com/moby/moby/issues/5899#issuecomment-54107108)
- https://github.com/sequenceiq/docker-pam/
- https://github.com/sequenceiq/docker-pam/blob/master/ubuntu-14.04/Dockerfile

It maybe a lot of work to remove the pi-hole references, if we can get the size down then this PR isn't as contentious for me.

dschaper:
Okay, my plan is to borrow some of the priv functions from `get.docker.com` and try to remove the `sudo` calls with `sh -c` if the user is UID/GID 0.

diginc:
Since this proved it worked, I think I should do this the right way and split out the synology pam debian modification into it's own image which this image uses as it's basis.  This makes the build faster & allows anyone else who may need a pam w/disable audit as a debian image to re-use it.

For now I think I'll just do amd64 since I don't think synology provides docker for anything but their intel models.

Azzhoe:
@diginc so currently I am using the official pihole/pihole:synology. Do I need to redownload it again using pihole/pihole:development instead in order to stay up to date with the latest working version for my Synology NAS?

Issue #361, opened 2018-11-08
ARM dev version errors with `exec user process caused "exec format error"`


diginc:
Full message : `standard_init_linux.go:190: exec user process caused "exec format error"`

With the recent `:dev` tag release Raspberry Pi 2+ kernels with `uname -m` armv7l+ versions can no longer run Pi-hole docker.  Note: Raspberry Pi 0/1 (armv6l) , errors are expected (usually no output).

Thanks for the report [in the other thread](https://github.com/pi-hole/docker-pi-hole/issues/336#issuecomment-436768162) @reloxx13 

```
$ docker run -it --rm pihole/pihole:v4.0_armhf
[s6-init] making user provided files available at /var/run/s6/etc...exited 0.
[s6-init] ensuring user provided files have correct perms...exited 0.
( works )

$ docker run -it --rm pihole/pihole:4.0.0-1_armhf
[s6-init] making user provided files available at /var/run/s6/etc...exited 0.
[s6-init] ensuring user provided files have correct perms...exited 0.
(works)

$ docker run -it --rm pihole/pihole:latest # (same as previous tag used but autodetecting)
[s6-init] making user provided files available at /var/run/s6/etc...exited 0.
[s6-init] ensuring user provided files have correct perms...exited 0.
(works)

$ docker run -it --rm pihole/pihole:dev_armhf
standard_init_linux.go:190: exec user process caused "exec format error"
```

So what changed?  Lets check the base image first off

```
$ docker history pihole/pihole:4.0.0-1_armhf | tail -3
<missing>           6 weeks ago         /bin/sh -c #(nop)  ENV ARCH=armv7l UBUNTU_SU…   0B
<missing>           6 weeks ago         /bin/sh -c #(nop)  CMD ["/bin/bash"]            0B
<missing>           6 weeks ago         /bin/sh -c #(nop) ADD file:fdfc0b2948d5940b5…   94.1MB
$ docker history pihole/pihole:dev_armhf | tail -3
<missing>           2 weeks ago         /bin/sh -c #(nop)  ENV ARCH=armv7l UBUNTU_SU…   0B
<missing>           2 weeks ago         /bin/sh -c #(nop)  CMD ["/bin/bash"]            0B
<missing>           2 weeks ago         /bin/sh -c #(nop) ADD file:abd539cf324221453…   94.2MB
```

Yup...so does that base image work by it's self?

```
docker run -it --rm multiarch/debian-debootstrap:armhf-stretch-slim date
[...]

PR #370, opened 2018-12-11
Version 4.1

diginc:
Here are code changes for the new version's release.  I'm running some more automated tests locally first.

I need to check the multi-architecture version out to see if it's still suffering from a problem that appeared after it's base image changed (see #361)


Dockerfile.template
```diff
@@ -34,7 +29,7 @@ EXPOSE 443
 ENV S6_LOGGING 0
 ENV S6_KEEP_ENV 1
 ENV S6_BEHAVIOUR_IF_STAGE2_FAILS 2
-ENV FTL_CMD no-daemon
+ENV FTL_CMD debug
```
diginc: @pi-hole/api-approvers FYI no-daemon wasn't staying in the foreground for me in my hasty update last night, so I changed it to debug as a workaround to get things running (s6-service requires a foregrounded process)

install.sh
```diff
@@ -1,22 +1,25 @@
-#!/bin/bash -ex
+#!/bin/bash -ex /
```
diginc: wait where did this / come from.......

install.sh
```diff
@@ -63,10 +67,12 @@ mv "${tmpLog}" /
 
 if [[ $USE_CUSTOM_BRANCHES == true ]] ; then
     ln -s /bin/true /usr/local/bin/service
-    echo "$CORE_TAG" | tee /etc/pihole/ftlbranch
+    ln -s /bin/true /usr/local/bin/update-rc.d
     echo y | bash -x pihole checkout core $CORE_TAG
     echo y | bash -x pihole checkout web $CORE_TAG
+    echo y | bash -x pihole checkout ftl ${CORE_TAG/v/}
```
diginc: New way to checkout dev copies of FTL added as I was pre-release testing.  Noticed no 'v' was in this tag where as the others weren't so I'm search/replacing it in this bash variable.

s6/debian-root/etc/cont-init.d/20-start.sh
```diff
@@ -10,6 +10,10 @@ fi
 # used to start dnsmasq here for gravity to use...now that conflicts port 53
 
 $bashCmd /start.sh
+# Gotta go fast, no time for gravity
```
diginc: Maybe not required, added before I added `--dns 1.1.1.1` which actually solved my gravity issues (known issue that never manifested in tests before)

test/test_bash_functions.py
```diff
@@ -18,7 +19,11 @@ def test_IPv6_not_True_removes_ipv6(Docker, args, expected_ipv6, expected_stdout
 
     function = Docker.run('. /bash_functions.sh ; setup_ipv4_ipv6')
     assert "Using {}".format(expected_stdout) in function.stdout
-    config = Docker.run('cat {}'.format(WEB_CONFIG)).stdout
+    if expected_stdout == 'IPv4':
+        assert 'IPv6' not in function.stdout
+    # weird slow write/sync problem; no sleep == old state of file, sleep == updated/setup state of file
+    time.sleep(1)
```
diginc: `setup_ipv4_ipv6` does what it is supposed to do according to every manual test I perform but automation/docker seems to read the file from the old state unless you wait a small amount of time at which point

tox.ini
```diff
@@ -4,10 +4,10 @@ envlist = py27
 [testenv]
 whitelist_externals = docker
 deps = -rrequirements.txt
-# auto parallel max b/c race condition with docker fixture (I think?)
+# 2 parallel max b/c race condition with docker fixture (I think?)
```
diginc: My local > 2 core machine started failing more when I set this to auto.  I thought I had a parallel race condition fixed previously but there maybe more or perhaps it's related to the `time.sleep()` "fix"; slow to sync/write disks in docker's overlay2 FS.

diginc:
Hopefully my commentary helps other understand more of these changes where I didn't leave enough code comments.

I pushed the [dev image tag](https://hub.docker.com/r/pihole/pihole/tags/) with 4.1 which needs a little more testing before going to the prod tag.

Besides changing FTL `no-daemon` to `debug` the test's break/fixes had little to do with the updating of pi-hole it's self and more to do with existing/known flakiness in tests.

Dockerfile.template
Mcat12: Thanks, I found the issue: pi-hole/FTL#430

DL6ER: I just tried to run it as `pihole-FTL no-daemon` and - as @Mcat12 commented in his PR - observed that Ctrl+C, etc. are not caught anymore, however, `pihole-FTL` indeed stays in teh foreground and works as expected. We should continue this here: https://github.com/pi-hole/FTL/issues/432

diginc:
The dev tag was suffering from an issue I had encountered previously with 'development' - the fix was `rm -r ~/.docker/manifests/<image>` to properly clear out that manifest's old versions.

This should be good to go, I tried the new features on my copy and test are passing-ish (albeit flakily, I have a separate plan for that).

diginc:
As I feared the arm images still give an error when ran on real arm hardware.  The multiarch base images aren't good anymore :( I'll have to dig into this more and report back.  What I tried last time is peeking at the history of our current v4.0 image to see if I could pull out the base layer of the old version of multiarch (they don't seem to have any old tags) but the history command just says "missing" instead of giving a hash to tag.

nemchik:
Any chance you guys might be willing to have a chat with the team at linuxserver.io to discuss the build process for this image? I asked if any of them might have any insight on this PR and they are open to discussing it on their discord server. To any who don't know lsio is a team of pretty awesome people who build a lot of pretty popular docker images.

diginc:
Found the root of the issue : https://github.com/pi-hole/docker-pi-hole/pull/358/commits/10c44c9374e5541e0f0e5c00729444799b6e4933

I had introduced a template variable to make updating s6 a single change rather than 3 and somehow it got reverted / hardcoded again during the #358 work.

Travis is being very slow to run but tests are passing very consistently locally now so I think I'll force this through and release.

pralor:
This pull request has been mentioned on **Pi-hole Userspace**. There might be relevant details there:

https://discourse.pi-hole.net/t/4-1-docker-logging-to-stdout/15633/5

PR #380, opened 2019-01-01
Disable stdout for docker logging of pihole-FTL

diginc:
This is part 1/2 to reduce logging & be more courteous to our arm users who have SD cards that have limited lifespan of I/O.  We still need to get pihole-FTL no-daemon mode working since the same debug output goes to /var/log/pihole-FTL.log.  Related : https://github.com/pi-hole/FTL/issues/432

This doesn't remove anything but de-duplicates output.  Instead of seeing the dnsmasq/FTL output in `docker logs` simply use /var/log/pihole.log and /var/log/pihole-FTL.log now.


Dockerfile.template
```diff
@@ -29,7 +29,7 @@ EXPOSE 443
 ENV S6_LOGGING 0
 ENV S6_KEEP_ENV 1
 ENV S6_BEHAVIOUR_IF_STAGE2_FAILS 2
-ENV FTL_CMD debug
+ENV FTL_CMD no-daemon
```
DL6ER: I'd say that this is actually the main change here

diginc: Earlier comment was Part 1 main change

This is definitely part 2 main change.

test/conftest.py
```diff
@@ -8,7 +8,7 @@
 def DockerGeneric(request, args, image, cmd, entrypoint=''):
     assert 'docker' in check_output('id'), "Are you in the docker group?"
     if 'pihole' in image:
-       args += " --dns 127.0.0.1 --dns 1.1.1.1 -v /dev/null:/etc/pihole/adlists.default -e PYTEST=1"
+       args += " --dns 127.0.0.1 --dns 1.1.1.1 -v /dev/null:/etc/pihole/adlists.default -e PYTEST=1 --cap-add=NET_ADMIN"
```
diginc: This is what enabled part 2's fix (debug -> no-daemon) to work.

The container failed pretty badly before adding, this capability didn't seem to be required in prior versions.

DL6ER:
Fixes https://github.com/pi-hole/FTL/issues/432

pralor:
This pull request has been mentioned on **Pi-hole Userspace**. There might be relevant details there:

https://discourse.pi-hole.net/t/4-1-docker-logging-to-stdout/15633/10

PR #383, opened 2019-01-10
Tidy up docker run script

vhp:
Cleanup trailing white-space, More considerate output for beginners and general cleanup around bash best practices. Plus we ensure the password is outputted for user. By default currently it's not and the user has to run the command manually again. 

## Description
The following does some general cleanup such as removing whitespace and ensuring bash best practices.  Most importantly though it Adds a wait so the container has some time to start up. When done the user will get the password they set, even if they hardcoded it into their script. 

The output will show the container is still starting with successive dots being printed until a container is marked as healthy. 

## Motivation and Context
Upon setting up pi-hole for myself I felt this script could have been easier to use for a beginner. I found myself having to dig for the random password

## How Has This Been Tested?
Use it myself. 

## Types of changes

## Checklist:

If you have any concerns let me know.

### DIFF
--- a/docker_run.sh
+++ b/docker_run.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bash -e
 # Lookups may not work for VPN / tun0
 IP_LOOKUP="$(ip route get 8.8.8.8 | awk '{for(i=1;i<=NF;i++) if ($i=="src") print $(i+1)}')"
 IPv6_LOOKUP="$(ip -6 route get 2001:4860:4860::8888 | awk '{for(i=1;i<=NF;i++) if ($i=="src") print $(i+1)}')"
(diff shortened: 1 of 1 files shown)


diginc:
I added a `-e` to help prevent entering the loop when a port or container name conflict is hit.  

It might be good to have another condition besides healthy on the loop check, such us unhealthy/restarting or a timeout, in case the user has start up script problems which cause the container to never be healthy.

Other than that I like it.

vhp:
I will be following up with a second PR to do your recommendations as 
well. I was just a tad busy today.

Thanks

diginc:
Already made some changes for limiting it to a set amount of checks : https://github.com/pi-hole/docker-pi-hole/pull/386/files#diff-021bfe9ad57343db93d07b1f3a38c24dR33

PR #390, opened 2019-01-13
Extend sleep and add env file for usage

vhp:
The previous changes we made were a nice step forward. But I found that with the 1 second sleep we could fail but still succeed if that makes sense. We would hit 20 iterations but it would all still work and tell the user it failed. This extends the sleep to 3 seconds and adds use of the .env file. We also update the Readme. 

## Description
Pretty much the same as above. Though I updated the Readme which was forgotten last time. The env file is a nice addition I think. Some companies and people (like myself) will want to set passwords or use their automation to template the .env file. One major positive for the .env file is for us to safely dev in the same branch we run out of. Not defining WEBPASSWORD certainly does that but it is nice to have a consistent personal password so you can access pihole from a phone for example to whitelist a site. We also add the .env file to the .gitignore so it's not accidentally committed. 

## Motivation and Context
Upon testing our last collaboration I noticed that 20 seconds of sleep was close to adequate but occasionally on my rasp pi 3 b+ i hit 20 seconds. Given that I moved the sleep to 3 seconds per iteration. This averages around 10 for me in testing which is pretty good as a mid point. 

## How Has This Been Tested?
I am testing in my home and things seem Great

## Types of changes

## Checklist:

Thanks. Happy to make any changes you suggest. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,6 @@ Here is a rundown of the other arguments passed into the example `docker run`:
 
 | Docker Arguments | Description |
 | ---------------- | ----------- |
-| `--env-file .env` <br/> *Optional* | File to store environment variables for docker. Here for convenience
 | `-p 80:80`<br/>`-p 53:53/tcp -p 53:53/udp`<br/> **Recommended** | Ports to expose, the bare minimum ports required for Pi-holes HTTP and DNS services
 | `--restart=unless-stopped`<br/> **Recommended** | Automatically (re)start your Pi-hole on boot or in the event of a crash
 | `-v /dir/for/pihole:/etc/pihole`<br/> **Recommended** | Volumes for your Pi-hole configs help persist changes across docker image updates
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -62,8 +64,23 @@ docker run -d \
     --dns=127.0.0.1 --dns=1.1.1.1 \
     pihole/pihole:latest
 
-echo -n "Your password for https://${IP}/admin/ is "
-docker logs pihole 2> /dev/null | grep 'password:'
+
+printf 'Starting up pihole container '
```
diginc: I don't want this much arbitrary bash on the first page of the readme.  I don't mind hiding it in the script.  At this point I feel like instead of mirroring the contents of docker_run.sh we just link to it.

vhp:
Good Idea. I went ahead and tidied it up.

PR #394, opened 2019-01-17
Update with 4.1.1+ requirement for cap_add

chapb:
## Description

Uncommented the line including cap_add, as it's suggested/required in 4.1.1+ to start DNS service, not just for DHCP. Moved the latter portion of the description lower.

## Motivation and Context

This previously optional argument is now required or strongly encouraged as per the README, so I suggest it be included by default in the example compose. Additionally, it may be worthwhile to highlight in the documentation somewhere that the argument cap-add is cap_add in compose. Caused me a bit of (inexperienced) headache.

## Types of changes

## Checklist:

### DIFF
--- a/doco-example.yml
+++ b/doco-example.yml
@@ -10,9 +10,9 @@ services:
       - "53:53/udp"
       - "80:80/tcp"
       - "443:443/tcp"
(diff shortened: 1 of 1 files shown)


diginc:
Good catch!  Thanks.

PR #413, opened 2019-02-04
v4.2.1 & more robust user root fix

diginc:
Gets rid of the 'update available' flashing in the footer and handles the tweak to dnsmasq's user in 4.2, 4.2.1 and legacy configs users may have volume mounted.


mhofman:
@diginc please see my suggestion on how to avoid running as root:
https://github.com/pi-hole/FTL/issues/432#issuecomment-460723702

Issue #392, opened 2019-01-14
Issue with running Pi-hole 4.1.1 in swarm (--cap-add)


shanekorbel:
**In raising this issue, I confirm the following:** `{please fill the checkboxes, e.g: [X]}`

- [X] I have read and understood the [contributors guide](https://github.com/pi-hole/pi-hole/blob/master/CONTRIBUTING.md).
- [X] The issue I am reporting can be *replicated*.
- [X] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/pi-hole/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/pi-hole/issues)).

**How familiar are you with the the source code relevant to this issue?:**

`{1}`

---
**Expected behaviour:**

`{Up until 4.1.1 I have been running pihole as a docker swarm service on a x64 linux cluster running ubuntu 16.04 (five node swarm). This required a little modifying of the normal install, but only in how to start up the image, no real functionality changes. This has allowed me to scale pihole to 2 replicas for redundancy and be able to allow pihole to run on any of the five hosts in the swarm should one fail. 

 }`

**Actual behaviour:**

`{Once i pulled the latest image (4.1.1) my service would no longer boot at all. I even removed my service and reset it up thinking it was something I accidently changed since i was trying to also configure placement preferences. Eventually it led me to the image repository and i noticed changes were made that required flags be added that are not allowed in swarm. 

> Starting with the v4.1.1 release your Pi-hole container may encounter issues starting the DNS service unless ran with the following settings:
> 
> --cap-add=NET_ADMIN This previously optional argument is now required or strongly encouraged
> Starting in a future version FTL DNS is going to check this setting automatically
> --dns=127.0.0.1 --dns=1.1.1.1 The second server can be any DNS IP of your choosing, but the first dns must be 127.0.0.1
> A WARNING stating "Misconfigured DNS in /etc/resolv.conf" may show in docker logs without this.
> These are the raw docker run cli versions of the commands. We provide no official support for docker GUIs but the community forums may be able to help if you do not see a place for these settings. Remember, always consult your manual too!  

}`

**Steps to reproduce:**

`{I realize this is somewhat of a special config since running pihole as docker service is not the way it is typically ran, however swarm and compose are very similar in most regards other then the issue i found...

[...]

PR #415, opened 2019-02-06
Allow running pihole-FTL as non-root

### LINKED ISSUE #414: Explore setcap on FTL possibilities
Making a new issue to discuss @mhofman's [suggestion to setcap inside the image build](https://github.com/pi-hole/docker-pi-hole/pull/413#issuecomment-460726178)

> Here is a summary of what I'm proposing:
> 
> Now:
> 
>   * Change docker image to do `setcap CAP_NET_BIND_SERVICE,CAP_NET_RAW,CAP_NET_ADMIN+ei usr/bin/pihole-FTL` on first container launch.
>   * Change docker image to drop root before launching the `pihole-FTL` process. The permitted capabilities of the process will be whatever the intersection is between the container's capabilities and the 3 capabilities above. By default this will be `CAP_NET_BIND_SERVICE,CAP_NET_RAW` unless the user launches the container with `cap_add: net_admin`.
>   * Check at runtime (either in pihole-FTL or in script) which capabilities are available, and warn user if not all 3 are
>
> In the future:
> 
>   * Improve the warning above (or make it an error) if the dnsmasq/pihole configuration uses a feature requiring a missing capability (dhcp-range without dhcp-broadcast -> need net_admin, dhcp-range without no-ping -> need net_raw, possibly others, I haven't verified all combinations)
> 
>   * Very far in the future, once you're really sure about which configuration causes which capabilities to be needed, drop all unnecessary capabilities.

Great ideas, really.  My first thought was this could work but then I recalled `docker build` has to support capabilities / setcap / privileged :(

https://github.com/moby/moby/issues/1916 - I'll have to read this more in full after work but this caught my eye:
- [workaround](https://github.com/moby/moby/issues/1916#issuecomment-335369216)

Just when I was starting to like my current build process ... would this be a big enough QoL improvement for users to justify such an weird build?  Or should runtime be explored as a possibility.

I'm honestly considering making the default docker run obsolete and using docker-compose as the default...which could potentially enable doing something like a privileged sidecar to handle setcap (I'm opposed to long running overly privileged containers).

mhofman:
## Description
- Add a new container environment variable allowing to specify the user to run the pihole-FTL process as. Defaults to root.
- Set inherited capabilities attributes on the pihole-FTL file to automatically grant runtime permitted capabilities when available in the bounding set. This allows dropping root before starting pihole-FTL without failing with a permission error if the capabilities are not available to the container (the process may still error out if performing an operation requiring the capability).
- Add some information on capabilities to the Readme file.

## Motivation and Context

Running FTLDNS as non-root is recommended for security reasons.
It also causes dnsmasq to skip the capabilities syscalls, allowing the container to run with more limited capabilities (i.e. ability to not add NET_ADMIN, or even drop NET_RAW).
Fixes #414, #392.

## How Has This Been Tested?

Ran the automated tests but my development environment seem to have issues running them even before changes.
To test the changes I performed the following manual tests:
- set `DNSMASQ_USER=root`, use `cap-add=NET_ADMIN`. Verify process is still run as root with all capabilities from the bounding set as effective.
- set `DNSMASQ_USER=pihole`, use `cap-add=NET_ADMIN`. Verify process is run as `pihole` with the NET_BIND_SERVICE, NET_RAW and NET_ADMIN capabilities as effective. Verify that the container starts successfully and that FTLDNS answers queries.
- set `DNSMASQ_USER=pihole`, use `cap-drop=NET_RAW`. Verify process is run as `pihole` with only the NET_BIND_SERVICE capability as effective. Verify that the container starts successfully and that FTLDNS answers queries.
- set `DNSMASQ_USER=root`, use `cap-drop=NET_SETFCAP`. Verify that container starts and process runs as root. This is to simulate a failure in setting the file capabilities attributes, which doesn't matter if the process is run as root.
- set `DNSMASQ_USER=pihole`, use `cap-drop=NET_SETFCAP`. Verify container fails checks and outputs an error message explaining the container cannot run as non-root.

I guess those could be automated, but I'm really not familiar enough with python for that.

## Types of changes

## Checklist:


mhofman:
This change should also fix issues such as #392 without reverting to the `no-daemon` hack:
when the user is not root, dnsmasq will not try to grab the net_admin (or other) capabilities.

dschaper:
LGTM

jacobsalmela:
@mhofman please contact me directly for a special thank you.

PR #417, opened 2019-02-08
Enhance systemd support and document it

trini:
## Description
- Since we encourage throw-away containers, document having the systemd unit operate in that manner.
- In addition, have the systemd unit determine the IPv4 and IPv6 server addresses dynamically.

## Motivation and Context
The README discourages the use of long-living containers, switching the systemd unit to one that removes itself on exit moves in that direction.  In addition, it can take a little digging to find a good example for "How do I find my IP, dynamically?" so populating that logic here will save folks time.

## How Has This Been Tested?
I've been running this unit (with a few more advanced options passed in too) for some time.

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,9 +148,9 @@ Similarly for the webserver you can customize configs in /etc/lighttpd
 
 ### Systemd init script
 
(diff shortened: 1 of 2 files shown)


diginc:
I pushed a incomplete update of ServerIP to the `dev` branch last night where I've removed the requirement of ServerIP and as a result simplified the run commands.   You should be able to remove the IP lookup logic all together, also please switch your PR's destination branch to 'dev'.    https://github.com/pi-hole/docker-pi-hole/pull/418

trini:
OK, I'll get a new one up shortly.

PR #420, opened 2019-02-08
Enhance systemd support and document it

trini:
## Description
- Since we encourage throw-away containers, document having the systemd unit operate in that manner.

## Motivation and Context
The README discourages the use of long-living containers, switching the systemd unit to one that removes itself on exit moves in that direction.

## How Has This Been Tested?
This unit, when based on current master is something (with a few more advanced options passed in too) for some time.

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,9 +187,9 @@ Similarly for the webserver you can customize configs in /etc/lighttpd
 
 ### Systemd init script
 
-As long as your docker system service auto starts on boot and you run your container with `--restart=unless-stopped` your container should always start on boot and restart on crashes.  If you prefer to have your docker container run as a systemd service instead, add the file [pihole.service](https://raw.githubusercontent.com/pi-hole/docker-pi-hole/master/pihole.service) to "/etc/systemd/system"; customize whatever your container name is and remove `--restart=unless-stopped` from your docker run.  Then after you have initially created the docker container using the docker run command above, you can control it with "systemctl start pihole" or "systemctl stop pihole" (instead of `docker start`/`docker stop`).  You can also enable it to auto-start on boot with "systemctl enable pihole" (as opposed to `--restart=unless-stopped` and making sure docker service auto-starts on boot).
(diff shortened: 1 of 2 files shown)


diginc:
@middleagedman since you wrote the original systemd stop/start version could you review/approve?

middleagedman:
From a logic standpoint yea, it somewhat makes sense.  However I feel dirty for some reason seeing a docker run command in a systemd exec statement.. Not sure why.  

It's probably fine from a dev perspective.  Some reasons (which could affect developers if they aren't savvy)
- Users may lose all pihole log data between reboots
- Having to keep the systemd service file maintained better.

Other than that.. I don't have any good arguments against it :)

middleagedman:
PS - I haven't been keeping up with the dev changes.. Do the docker builds no longer require the following?
-p 53:53/tcp -p 53:53/udp
-e VIRTUAL_HOST=
-e TZ=
-e WEBPASSWORD=
-v /etc/localtime:/etc/localtime:ro

trini:
> PS - I haven't been keeping up with the dev changes.. Do the docker builds no longer require the following?
> -p 53:53/tcp -p 53:53/udp
> -e VIRTUAL_HOST=
> -e TZ=
> -e WEBPASSWORD=
> -v /etc/localtime:/etc/localtime:ro

I went with --net host rather than spelling out the ports (as we'd also need to pass in 80) and the rest are either optional or not required now, per the README.md at least.

trini:
> From a logic standpoint yea, it somewhat makes sense. However I feel dirty for some reason seeing a docker run command in a systemd exec statement.. Not sure why.

I'm open to suggestions.  But part of the point is to use run and --rm so we get a new container each time.  I didn't see a better way to get that, but I'm not a Docker expert, just a user.

> It's probably fine from a dev perspective. Some reasons (which could affect developers if they aren't savvy)
> 
>     * Users may lose all pihole log data between reboots
> 
>     * Having to keep the systemd service file maintained better.
> 
> 
> Other than that.. I don't have any good arguments against it :)

Both of these are true.  For the former, we could add in words about a volume being used for logs and thus they become really persistent rather than persistent until re-created.

middleagedman:
@diginc If we approve this in dev will it get auto-promoted to prod code eventually?

trini:
As the rest of the README as moved in another direction, while the contents here apply I don't think it fits with the _spirit_ at this point, so closing.

Issue #443, opened 2019-04-19
Unable to run pi-hole with network mode "host" on a Raspberry 


nikoskip:
This is a... 

- [ ] Request for a new or modified feature
- [x] Issue trying to run the docker image
- [ ] Issue trying to build / test / develop the docker image

## Description
Pi-hole is unable to start due a httpd port bind issue.

## Expected Behavior
Start Pi-hole normally

## Actual Behavior
After running `docker-compose up` I'm getting this message on screen:

```
pihole    | 2019-04-18 20:28:26: (network.c.464) can't bind to port: 127.0.0.1 80 Address already in use
pihole    | Stopping lighttpd
pihole    | lighttpd: no process found
pihole    | Starting lighttpd
pihole    | 2019-04-18 20:28:27: (network.c.464) can't bind to port: 127.0.0.1 80 Address already in use
pihole    | Stopping lighttpd
pihole    | lighttpd: no process found
pihole    | Starting lighttpd
pihole    | 2019-04-18 20:28:28: (network.c.464) can't bind to port: 127.0.0.1 80 Address already in use
pihole    | Stopping lighttpd
pihole    | lighttpd: no process found
pihole    | Starting lighttpd
pihole    | 2019-04-18 20:28:29: (network.c.464) can't bind to port: 127.0.0.1 80 Address already in use
pihole    | Stopping lighttpd
```

This only happens using **host** as network mode on Docker, and I would like to use the host mode because I need the DHCP functionality.

## Possible Fix
I don't know 😢 

## Steps to Reproduce and debugging done
Use this `docker-compose.yml` and try to run it on a Raspberry Pi:

```
version: "3"

# More info at https://github.com/pi-hole/docker-pi-hole/ and https://docs.pi-hole.net/
services:
  pihole:
    container_name: pihole
    image: pihole/pihole:latest
    environment:
      TZ: 'America/Santiago'
      WEBPASSWORD: 'xxxx'
    # Volumes store your data between container upgrades
    volumes:
       - './etc-pihole/:/etc/pihole/'
       - './etc-dnsmasq.d/:/etc/dnsmasq.d/'
    dns:
      - 127.0.0.1
      - 1.1.1.1
    # Recommended but not required (DHCP needs NET_ADMIN)
    #   https://github.com/pi-hole/docker-pi-hole#note-on-capabilities
    cap_add:
      - NET_ADMIN
    restart: unless-stopped
    network_mode: host
```

## Debug steps I have tried

[...]

Issue #432, opened 2019-03-02
Lighttpd doesn't start on fresh container with --net=host


savef:
Hiya,

This is a... 

- [ ] Request for a new or modified feature
- [x] Issue trying to run the docker image
- [ ] Issue trying to build / test / develop the docker image

## Description

I have been trying to use docker-compose to launch a container, and it gets stuck in a loop of trying and failing to bind Lighttpd. Here is my config yaml file:
```
version: "3"
services:
  pihole:
    container_name: pihole
    image: pihole/pihole:latest
    network_mode: host
    environment:
      TZ: 'Europe/London'
      WEBPASSWORD: 'helloW0rld' # temporary
    volumes:
       - './etc-pihole/:/etc/pihole/'
       - './etc-dnsmasq.d/:/etc/dnsmasq.d/'
    dns:
      - 127.0.0.1
      - 1.1.1.1
    # Recommended but not required (DHCP needs NET_ADMIN)
    #   https://github.com/pi-hole/docker-pi-hole#note-on-capabilities
    cap_add:
      - NET_ADMIN
    restart: unless-stopped
```
I also tried changing WEB_PORT to 1080, the port variable is being pulled through but the same error loop massage appears:
```
pihole    | Starting lighttpd
pihole    | 2019-03-02 14:19:05: (network.c.464) can't bind to port: 127.0.0.1 1080 Address already in use
pihole    | Stopping lighttpd
pihole    | lighttpd: no process found
pihole    | Starting lighttpd
pihole    | 2019-03-02 14:19:06: (network.c.464) can't bind to port: 127.0.0.1 1080 Address already in use
pihole    | Stopping lighttpd
pihole    | lighttpd: no process found
```
Nothing is running on these ports on the host, as confirmed by netstat:
```
Here's the full `docker-compose up` output: [https://pastebin.com/XX8kf36h](https://pastebin.com/XX8kf36h)

~/containers/pi-hole$ sudo netstat -ltpn
Active Internet connections (only servers)
Proto Recv-Q Send-Q Local Address           Foreign Address         State       PID/Program name
tcp        0      0 0.0.0.0:22              0.0.0.0:*               LISTEN      1174/sshd
tcp6       0      0 :::22                   :::*                    LISTEN      1174/sshd
```

## Debug steps I have tried

- [x] I have tried destroying my container instance, pulling the newest image version, and re-creating a new container - no effect
[...]

PR #447, opened 2019-04-24
Update README.md

diginc:
Clarify that people running --net=host should set ServerIP for Pi-hole to function correctly.

I'll have to try to add some automation around this later to try to warn only the network host mode users if they miss it.

See : #432 #443

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ services:
 
 ### Docker Pi-Hole v4.2.2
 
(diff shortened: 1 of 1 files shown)


itcotta:
It should perhaps be also worth noting that the net-host only works in linux hosts.
Windows running docker doesn't support it according to: https://docs.docker.com/network/host/

> The host networking driver only works on Linux hosts, and is not supported on Docker Desktop for Mac, Docker Desktop for Windows, or Docker EE for Windows Server.

PR #451, opened 2019-05-08
Use volume for configuration directory

mabunixda:
## Description
Dockerfile sets /etc/pihole as volume and provides this information as environment variable. 
Within this directory now also the first boot check is pending on either the directory is empty - in case of using k8s and a volume mount - or the /etc/pihole/.piholeFirstBoot is still available.

## Motivation and Context
Because I started running 2 pihole instances within k8s I encountered the problem that /.piholeFirstBoot is always available and the password is reset all the time. 

## How Has This Been Tested?
K8s run is tested within my local k8s cluster, docker run is tested locally on a test system and test cases still work ( local + [travis](https://travis-ci.org/mabunixda/docker-pi-hole/branches)

## Types of changes

## Checklist:


diginc:
Looks good to me, merged to dev to make a test image.

PR #462, opened 2019-05-31
Fix to docker-compose-jwilder-proxy.yml formatting

mvaal:
Running the example docker-compose-jwilder-proxy.yml using version 3 of docker-compose was throwing exceptions. This changes corrects the example.

## Description
Was getting the below error when attempting to run the example docker-compose-jwilder-proxy.yml file.
```
Additional properties are not allowed ('pihole', 'applist' were unexpected)

You might be seeing this error because you're using the wrong Compose file version. Either specify a supported version ("2.0", "2.1", "3.0", "3.1", "3.2") and place your service definitions under the `services` key, or omit the `version` key and place your service definitions at the root of the file to use version 1.
For more on the Compose file format versions, see https://docs.docker.com/compose/compose-file/
```

## Motivation and Context
Help future people use the example.

## How Has This Been Tested?
Ran a local copy with the corrected changes.

## Types of changes

## Checklist:

### DIFF
--- a/docker-compose-jwilder-proxy.yml
+++ b/docker-compose-jwilder-proxy.yml
@@ -2,61 +2,62 @@ version: "3"
 
 # https://github.com/pi-hole/docker-pi-hole/blob/master/README.md
 
-applist:
-  image: jwilder/nginx-proxy
-  ports:
-    - '80:80'
-  environment:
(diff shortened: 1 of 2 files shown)


mvaal:
Accidentally pulled is some other changes somehow, will resubmit.

PR #472, opened 2019-06-01
Revert 451 k8s

diginc:
I tried to implement #451 another way quickly but I'm out of time for today so I'm just going to revert it for now.


diginc:
FYI I haven't given up on you @mabunixda or #451 - https://github.com/pi-hole/docker-pi-hole/tree/support_k8s_volumes

I started by just renaming and refactoring some of the original first container start / initial setup logic to my liking and ended up breaking some stuff.  I'll have to dig into this code more a little later.

PR #487, opened 2019-08-05
Change default DNS provider to Cloudflare

martinjuhasz:
This will change the default settings for the DNS provider from Google to Cloudflare.

## Description
All occurences of the DNS settings for Primary and redundancy have been changed to `1.1.1.1` and `1.0.0.1`.

## Motivation and Context
Pi-hole is all about privacy. Still the default DNS provider that will be used is Google's. Google logs quite some things when you use their DNS service like the IP address (temporary) but also a lot of things permanently, like: geolocation information, request domain name, request type and [a lot more](https://developers.google.com/speed/public-dns/privacy).

I suggest switching the default DNS provider to[ Cloudflares 1.1.1.1 service](https://1.1.1.1/dns/), as their mantra is "Privacy First: Guaranteed" and is also [proven faster](https://www.dnsperf.com/#!dns-resolvers).

> We will never log your IP address (the way other companies identify you). And we’re not just saying that. We’ve retained KPMG to audit our systems annually to ensure that we're doing what we say.

## Types of changes

## Checklist:


dschaper:
This isn't a change that should be made in the docker image. This is a change, that if approved, needs to be made in the core project, https://github.com/pi-hole/pi-hole.

martinjuhasz:
The main project is agnostic and doesnt have something like default dns providers right, so its really just about the docker i guess?

> This is a configuration that can be overridden by the user. I really see no need to change anything nor make a statement as to the preferred default DNS provider.

I see your point, still it's the default configuration. A lot of people use pi-hole, and a lot of them don't even know what that DNS setting is for, since this is not only a tool for developers or tech persons, but even non tech people tend to fire up a Pi with this, to get rid of ads. I'm just proposing a way of making pi-hole more privacy friendly **by default**, without additional configuration required.

dschaper:
This will be a problem as the IP address for CloudFlare is still being used by a large section of router hardware for internal use. If the change is becuase Google == Bad then I'll say that CloudFlare isn't to be trusted. They are for-profit, are going to IPO in a bit and have a CEO that makes major decisions based on how he felt when he woke up in the morning.

If you want to make changes then look for something like the non-profit Quad9 upstream.

martinjuhasz:
I see your point. Let's settle this then for now. I guess such a discussion would then be best placed in the main repos issues.

eku:
Cloudflare and privacy? Good joke!

PR #493, opened 2019-08-15
docker-compose cleanup

WithoutCaps:
If you guys are fine with these changes - later on I will submit commit a which fixes documentation

## Description
1. Removed `container_name` attribute since it literally does nothing in this case
2. Removed `TZ` attribute since this makes this docker-compose file location-specific. Usually, timezone automagically is set to the same one as host machine
3. Uncommented `WEBPASSWORD` environment variable, this way it is not set to anything unless a person has an environment variable set, this makes this file more flexible
4. Added "named volumes", this way pihole will store its data to the same location as all other dockers containers. This is a breaking change, it is not compatible with older versions of this file

## Motivation and Context
I am in the process of building a new home server and noticed that this file was quite nasty, so here we are...

## How Has This Been Tested?
Tested with `docker-compose` and `docker stack`, both seem to work just fine

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,29 +4,29 @@ version: "3"
 
 services:
   pihole:
-    container_name: pihole
     image: pihole/pihole:latest
     # For DHCP it is recommended to remove these ports and instead add: network_mode: "host"
     ports:
+      - "443:443/tcp"
       - "53:53/tcp"
       - "53:53/udp"
       - "67:67/udp"
       - "80:80/tcp"
-      - "443:443/tcp"
     environment:
-      TZ: 'America/Chicago'
-      # WEBPASSWORD: 'set a secure password here or it will be random'
-    # Volumes store your data between container upgrades
+      WEBPASSWORD: ${PASSWORD}
     volumes:
-       - './etc-pihole/:/etc/pihole/'
-       - './etc-dnsmasq.d/:/etc/dnsmasq.d/'
-    # run `touch ./var-log/pihole.log` first unless you like errors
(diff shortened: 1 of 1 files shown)


diginc:
Thanks for the submission, some of these choices were made to help the example commands work for the most amount of people.  Let me explain the choices and the intention behind them.

1) pihole as the container name helps make support consistent, `docker exec` debug commands will be ran against a container named pihole 90% of the time rather than a new docker user having to reply "it said no container found named pihole" because it was actually named something random (in docker run's case) or have a project folder name prefix (in docker-compose's case)

2) Volume mapping in the host's timezone is different on an OS basis.  Think the big 3: windows, linux, mac.  The goal is to prevent people complaining the timezone is UTC in docker but not having to provide highly variable docker/docker-compose run commands.  TZ=<variable> provides a common interface that needs updating by the user.  There is good in bad in this design choice, it is still variable but it's consistent across OSs.  I'm open to another consistent solution that works for multiple OSs that doesn't require an entire helper function/library/framework to map in time.

3) If WEB_PASSWORD is empty, which would happen when the $PASSWORD variable is not set or is empty, then that disables all security which goes against the intention of security by default.  The commented out WEB_PASSWORD generates a random password (the most secure option) and passively advertises there is a feature to override it.

4) I like this one, however I think named volumes are better served for persisting data that doesn't need exposure to the user like a backend database.  We have configuration files that can be hand modified in an editor and then the container can be restarted to update the changes.  That's worth mapping/mounting into a host directory isn't it?

PR #501, opened 2019-09-02
Update README.md

TheTonon:
I just found some typos on the Readme and wanted to get it right.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,8 +24,8 @@ services:
       - "80:80/tcp"
       - "443:443/tcp"
     environment:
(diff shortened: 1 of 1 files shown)


diginc:
https://docs.docker.com/compose/compose-file/#environment shows 2 styles of environment blocks.

- Key value map/dictionary: `key: value`
- A list of strings with keyvalue separated by an equals: "- key=value"

You seem to be confusing the two and mixing them together.  I don't believe the original is incorrect.

PR #512, opened 2019-09-22
Add support for setting the admin password from a file

kaysond:
## Description

Added a `-r` test for the `$WEBPASSWORD_FILE` variable; if it is readable, contents are loaded into `$WEBPASSWORD` and operation continues as normal.

## Motivation and Context

For various reasons, it might be desirable not to have the admin password stored in the docker-compose or .env files (e.g. they are in revision control). This allows the password to be loaded from a file in the container, rather than directly from an environment variable. The file can be set up as another bind mount, for example, or via [docker secrets](https://docs.docker.com/engine/swarm/secrets/).

## How Has This Been Tested?

Tested by building the container on a Ubuntu 18.04.3 host, setting this env var via docker-compose file, then running the container and logging in to the admin page.

## Types of changes

- [ X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X ] My code follows the code style of this project.
- [X ] My change requires a change to the documentation.
- [X ] I have updated the documentation accordingly.


kaysond:
@diginc Following up to see if any issues are preventing you from merging?

lightswitch05:
@kaysond I did not see this pull request and implemented the same concept in #584 - which is now in the dev branch if you want to give it a try.

PR #533, opened 2019-11-15
Add to Ubuntu install section

warren-fisher:
Add info about restarting systemd-resolved

## Motivation and Context
I was in the process of installing this onto my Ubuntu server and ran into some conflicts on port 53. The documentation added recently to the README about systemd-resolved led me to the right place. I made the necessary changes, and it did not work. Examining the [basic install script](https://github.com/pi-hole/pi-hole/blob/master/automated%20install/basic-install.sh) at line 1540 I found that the install script restarted systemd-resolved. I did not realize that this was necessary, but this did lead me to fixing the issue. I believe it would make sense to mention restarting systemd-resolved is required (I think it is atleast) to help anyone who ran into the same problem as me.

## How Has This Been Tested?
No testing needed, just a documentation update.

## Types of changes

I don't think any of these apply, it is only an addition to the documentation.

## Checklist:


warren-fisher:
Not sure why the build failed, never used any form of CI before. I just assumed that the build would succeed since I only made changes to the README.

Issue #573, opened 2020-02-23
Release v4.3.3 Docker Image


parkr:
This is a... 

- [ ] Request for a new or modified feature
- [x] Issue trying to run the docker image
- [ ] Issue trying to build / test / develop the docker image

## Description

Attempting to pull `docker pull pihole/pihole:4.3.3` fails.

## Expected Behavior

Expected to get the latest release of PiHole, v4.3.3.

## Actual Behavior

```text
~$ docker pull pihole/pihole:4.3.3
Error response from daemon: manifest for pihole/pihole:4.3.3 not found: manifest unknown: manifest unknown
```

## Possible Fix

Build & publish pihole v4.3.3 Docker image to Docker Hub.

## Steps to Reproduce and debugging done

e.g. your docker run command, pages to visit, CLI commands you ran
1. `docker pull pihole/pihole:4.3.3`

## Debug steps I have tried

- [ ] I have tried destroying my container instance, pulling the newest image version, and re-creating a new container
- [ ] I have tried running the nearly stock `docker run` example in the readme (removing any customizations I added)
- [ ] I have tried running without my volume data mounts to eliminate volumes as the cause
- [x] I have searched this repository for existing issues and pull requests that look similar 

## Your Environment 

* Docker Host Operating System and OS Version:
* Docker Version:
* Hardware architecture:

dschaper:
That version doesn't exist. We haven't released anything for it.

compuguy:
@dschaper There is an 4.3.3 update for the admin/web interface though: https://github.com/pi-hole/AdminLTE/releases/tag/v4.3.3

dschaper:
I know, we haven't released the docker image for it. There's more changes that may be made.

dschaper:
> Shouldn't that be enough to push a docker 4.3.3 release?

It will be released when it's ready. And when it's released you'll be able to pull it.

dschaper:
As this isn't an bug report I'll close it out.

compuguy:
> It will be released when it's ready. And when it's released you'll be able to pull it.
I **completely** understand that. I was a bit more concerned since that minor release fixes a security vulnerability: https://nvd.nist.gov/vuln/detail/CVE-2017-8816

dschaper:
That's not the right CVE?

compuguy:
Your right, @dschaper. I guess I clicked on an older vulnerability that had a different year, but the same last 4 digits. I've edited my previous post to correct that.

[...]

PR #574, opened 2020-02-23
update web version to 4.3.3

PromoFaux:
Addresses #573


PromoFaux:
Actually.. scratch this, we'll do it with the 4.3.3 core release

PR #579, opened 2020-02-25
Improve healthcheck responsiveness

lightswitch05:
* Improve healthcheck responsiveness by disabling dig retry logic.
* Disable recursive queries for the healthcheck.
* Remove duplicate python import

## Description
By default, dig will try 3 times to get a response back. Each attempt defaults to 5 seconds. Before this change, a single docker healtcheck failure would really mean three failures and would take a total of 15 seconds before failing. By default, docker healthchecks will retry 3 times before considering a service unhealhy (with a 30 second interval). Combined with dig retries, this means it would take a total of 9 failed DNS responses before it considers the pihole to be unhealthy. Combining the retry between dig and docker, dig considers it a success if even 1/3 responses are received - and docker considers it a success if only 1/3 of those attempts are successful. I'm not great at math - and order does make a difference - but I think that means as long as 1/9th of these queries are being answered - then docker thinks its healthy. Anyways, long story sort, dig doesn't need to have its own retry logic since docker already has a configurable retry. I also disable recurse since the goal is to test this specific instance.

## Motivation and Context
I have not had issues with this. I'm just trying to get familiar with this project and this stood out to me as a small potential improvement. 

## How Has This Been Tested?

I tested this by running dig commands directly, as well as building the image locally. I was unable to figure out how to cause a running image to become unhealthy but remain running.

## Types of changes

## Checklist:

### DIFF
--- a/Dockerfile.py
+++ b/Dockerfile.py
@@ -17,7 +17,6 @@ Examples:
 """
 
 
-from docopt import docopt
 from jinja2 import Environment, FileSystemLoader
 from docopt import docopt
 import os
--- a/Dockerfile.template
+++ b/Dockerfile.template
@@ -44,6 +44,6 @@ LABEL image="{{ pihole.name }}:{{ pihole.version }}_{{ pihole.arch }}"
 LABEL maintainer="{{ pihole.maintainer }}"
 LABEL url="https://www.github.com/pi-hole/docker-pi-hole"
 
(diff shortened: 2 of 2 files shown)


lightswitch05:
🍾 🎉 😄

diginc:
Thanks for the PR. I wasn't aware this was a problem. Nice fix.

PR #607, opened 2020-05-03
Update docker-compose-traefik-proxy.md according to changes in traefik2.X 

chinskiy:
## Description
Hello!
I've tried to run an example from [`docker-compose-traefik-proxy.md`](https://github.com/pi-hole/docker-pi-hole/blob/master/docker-compose-traefik-proxy.md) and noticed that it works only with traefik1.X releases. Probably I've even not noticed this but docker-compose use `:latest` tag for `traefik` and now it's 2.2 version". So, for now, the example doesn't work.
I've to change configs to make it acceptable with traefik2.X releases.

## Motivation and Context

Due to backward-incompatible changes in traefik2.X([link](https://docs.traefik.io/migration/v1-to-v2/)) current example doesn't work.

## How Has This Been Tested?
On my Raspberry Pi.

## Types of changes

## Checklist:


diginc:
Thanks! I was meaning to get to this upgrade.  My one suggestion would be pin the image to `image: traefik:v2.2` to avoid repeating history.

diginc:
Using multiple sites I had issues with the rule you had since it conflicted.

```
      - "traefik.enable=true"
      - "traefik.http.services.pihole.loadbalancer.server.port=80"
      - "traefik.http.routers.pihole.rule=Host(`pihole.diginc.lan`)"
      - "traefik.http.routers.pihole.entrypoints=http"
      - "traefik.docker.network=discovery"
```

Is closer to the original v1 config I had.

chinskiy:
Sure
I've moved back network name `discovery` and pin `traefik` image

But
````
- "traefik.http.routers.pihole.rule=Host(`pihole.diginc.lan`)"
````
I've change on 
````
- "traefik.http.routers.pihole.rule=Host(`pihole.homedomain.lan`)"
````
Because whole file reference on `homedomain.lan`

diginc:
I also removed `- /dev/null:/traefik.toml` since the command syntax gets picked up fine without that.

I'm going to change to master & merge since this is documentation and doesn't require any docker builds on `:dev`.

Thanks for the help!

PR #634, opened 2020-06-05
Introduce additional docker tags for the debian version.

lightswitch05:
Introduce additional docker tags for the Debian version, allows building both stretch and buster images.

## Description

* Added new docker tag variations to specify the debian version ('stretch', and 'buster').
* Arch images are alway as specific as possible: pihole/pihole:master-amd64-stretch
* Multiarch images have both the specific debian version tags as well as the generic non-debian tags: pihole/pihole:master-stretch & pihole/pihole:master
* Currently, the non-specific tags point to the 'stretch' images. Eventaully it can be migrated to 'buster'.
* Use GitHub actions to do the builds. Although the script names include 'gh-actions' to differentiate them from the 'circle' scripts, there is zero logic that is specific to Github (ie. no Github environment variables).
* 'armhf:buster' & 'arm64:buster' has an issue with `ip route get`. I think the issue is related to 'qemu', but I'm not sure. Update the `validate_env` function to only use `ip route get` if `nc` reports something strange.

## Motivation and Context

* Opens a migration path for eventually having buster images being the primary Debian version and future debian versions
* Uses Github Actions which are a little friendlier and easier to find. Doesn't require a circle-ci account.

## How Has This Been Tested?

* Locally and Github Actions

## Types of changes

## Checklist:


lightswitch05:
Note: Some of the Buster arch images are failing tests. I'm not really sure why. I wanted to go ahead and open the PR to get some early feedback on the changes. If anyone else knows what is going on with the `armhf:buster` and `arm64:buster` images, it would be greatly appreciated!

lightswitch05:
The failure: `ServerIP Environment variable (1.2.3.4) doesn't appear to be a valid IPv4 address`

With error: `Unsupported setsockopt level=270 optname=11`

From a quick search, looks like it might be related to qemu?

lightswitch05:
Just realized my manifest tagging is all messed up. Apologies for all the activity here. Closing PR for now, I'll reopen it once I have manifest tagging figured out.

Issue #10, opened 2016-04-14
image is not available?


debiansid:
docker run -p 53:53/tcp -p 53:53/udp -p 80:80 --cap-add=NET_ADMIN -e piholeIP="$IP" --name pihole -d dockerhole_alpine
Unable to find image 'dockerhole_alpine:latest' locally
Pulling repository docker.io/library/dockerhole_alpine
docker: Error: image library/dockerhole_alpine not found.
See 'docker run --help'.

diginc:
Sorry, that was the name of my docker-compose local build  it should be diginc/pi-hole for the image.  Updating docs now.

PR #635, opened 2020-06-07
Introduce additional docker tags for the debian version.

lightswitch05:
Introduce additional docker tags for the Debian version, allows building both stretch and buster images.

## Description

* Added new docker tag variations to specify the Debian version (`stretch`, and `buster`).
* **Arch images are always as specific as possible: `pihole/pihole:master-amd64-stretch`. This is a bit of a breaking change, as there are no more `pihole/pihole:master-amd64` tags... not sure if that is a problem?**
* Multiarch images have both the specific Debian version tags as well as the generic non-debian tags: `pihole/pihole:master-stretch` & `pihole/pihole:master`
* Currently, the non-specific tags point to the 'stretch' images. Eventually it can be migrated to `buster`.
* Use GitHub actions to do the builds. Although the script names include 'gh-actions' to differentiate them from the 'circle' scripts, there is zero logic that is specific to Github (ie. no Github environment variables).
* `armhf:buster` & `arm64:buster` has an issue with `ip route get`. I think the issue is related to `qemu`, but I'm not sure. Update the `validate_env` function to only use `ip route get` if `nc` reports something strange.
* Delete circle-ci job since it will fail with these changes. I've not included that here. I can delete the files if wanted.
* This introduces `buster` and `stretch` tags, other then getting them to build and pass tests, I haven't done any further verification of the `buster` images

## Motivation and Context

* Opens a migration path for eventually having buster images being the primary Debian version and future debian versions
* Uses Github Actions which are a little friendlier and easier to find. Doesn't require a circle-ci account.

## How Has This Been Tested?

* Locally and Github Actions.
* See the images build by my Github actions: https://hub.docker.com/repository/docker/lightswitch05/pihole
* See my Github actions: https://github.com/lightswitch05/docker-pi-hole/runs/746208461?check_suite_focus=true

## Types of changes

## Checklist:


.github/workflows/test-and-build.yaml
```diff
@@ -0,0 +1,65 @@
...
+      - dev
+      - v*
+      - beta-v*
+      # TODO @lightswitch05 remove this!
+      - feature/introduce-debian-version-tags
+  pull_request:
+
+# TODO @lightswitch05 remove this!
+env:
+  DOCKER_HUB_REPO: lightswitch05
```
lightswitch05, line 15: obviously this and line #10 needs to be removed. I'm leaving it in here until at least an initial review since it makes development on this branch much easier on my repo.

lightswitch05:
I know there is a lot in here to review. Any initial thoughts? Is the lack of debian-generic architecture tags a blocker?

dschaper:
I haven't looked it over since I try to stay out of @diginc way with docker but I'll give it a view if I'm not stepping on any toes.

dschaper:
@diginc need any help?

bash_functions.sh
```diff
@@ -46,7 +46,7 @@ validate_env() {
     # Optional ServerIP is a valid IP
     # nc won't throw any text based errors when it times out connecting to a valid IP, otherwise it complains about the DNS name being garbage
     # if nc doesn't behave as we expect on a valid IP the routing table should be able to look it up and return a 0 retcode
-    if [[ "$(nc -4 -w1 -z "$ServerIP" 53 2>&1)" != "" ]] || ! ip route get "$ServerIP" > /dev/null ; then
+    if [[ "$(nc -4 -w1 -z "$ServerIP" 53 2>&1)" != "" ]] && ! ip route get "$ServerIP" > /dev/null ; then
```
diginc, line 49: > if nc doesn't behave as we expect on a valid IP the routing table should be able to look it up and return a 0 retcode

The comment states netcat can be unreliable so ip route is a fallback, so it should be ||, not && shouldn't it?

lightswitch05: no, if `nc` fails with an `||`, then `ip route get` will never run. `&&` means that both will run each time, and if both `nc` and `ip route get` fail, then we know for sure it is no good. If either of them pass, then we can be reasonably sure its valid.

lightswitch05: The better solution would be to figure out why `armhf:buster` & `arm64:buster` has an issue with `ip route get`. I spent some time on it, but when I starting running into things involving `qemu`, I gave up.

lightswitch05: The confusing part is the they are both negative tests, so `!=` && `!=`

gh-actions-vars.sh
```diff
@@ -0,0 +1,53 @@
...
+
+# @environment ${ARCH}                    The architecture to build. Defaults to 'amd64'.
+# @environment ${DEBIAN_VERSION}          Debian version to build. Defaults to 'stretch'.
+# @environment ${DOCKER_HUB_REPO}         The docker hub repo to tag images for. Defaults to 'pihole'.
+# @environment ${DOCKER_HUB_IMAGE_NAME}   The name of the resulting image. Defaults to 'pihole'.
+
+GIT_BRANCH=$(git rev-parse --abbrev-ref HEAD | sed "s/\//-/g")
+GIT_TAG=$(git describe --tags --exact-match 2> /dev/null || true)
+
+DEFAULT_DEBIAN_VERSION="stretch"
```
diginc, line 12: we have a sprawling amount of default variables / updates to make in different places when those defaults change.  We don't need it immediately but at some point, I'd like to centralize them with a very early export of defaults in a centrally included envs.sh that the pipeline uses before any builds/tests/uploads run.  variables in the pipeline would work as a central location too but consider executing issues executing locally may have when putting things in the pipelines. The version number could migrate to this environment export too potentially.

.github/workflows/test-and-build.yaml
lightswitch05: removed!

lightswitch05:
I removed the references to my repo. Thanks for taking the time to review this massive PR!

PR #649, opened 2020-07-10
Feature/self contained docker run

aviadra:
## Description
Makes it so the "docker_run.sh" script, is self contained and doesn't require the user to fill in so many blanks.

## Motivation and Context
This is to make the docker installation method, as smooth as the regular one. One can install by simply curling the script into the bash

## How Has This Been Tested?
Test on:

- My raspberry pi running ubuntu 20.04
- AWS amd64 instance of Ubuntu 20.04
- AWS arm instance of Ubuntu 20.04
- AWS amd64 instance of RHEL8.

The only thing I changed is that I pointed the install script to my repo and my branch, as this is not part of the main repo yet.

This is my first PR to free software, I hope you like it.


diginc:
I like the sentiment of helping users get going quickly and easily.  I'll discuss with the team before making a more complete reply but I'll drop my first reactions:

- I prefer docker-compose as a default option (it was moved to the top of readme a while back)
- I've been seeing more success and recommendations for people to just use macvlan to avoid issues to the point I'm thinking it should be default and if we have to bootstrap to help start the container, that should be used perhaps.

aviadra:
Thank you for the prompt response.
Looking forward to hearing from you again.

docker_run.sh
```diff
@@ -1,29 +1,99 @@
...
-    --dns=127.0.0.1 --dns=1.1.1.1 \
-    --restart=unless-stopped \
-    --hostname pi.hole \
-    -e VIRTUAL_HOST="pi.hole" \
-    -e PROXY_LOCATION="pi.hole" \
-    -e ServerIP="127.0.0.1" \
-    pihole/pihole:latest
-
-printf 'Starting up pihole container '
+#!/usr/bin/env zsh
```
dschaper, line 1: Use bash or pure shell.

dschaper (changes requested):

dschaper:
>     * I prefer docker-compose as a default option (it was moved to the top of readme a while back)

Same here. Other options are environmental variables, [envdir](http://manpages.ubuntu.com/manpages/trusty/man8/envdir.8.html) is great for this.

aviadra:
Thank you. I was developing on a mac... I'll change it

lightswitch05:
Do we really want to encourage more `curl | bash` type behavior?

Beyond that, this script is now making changes to the host's `systemd-resolved` configuration. I understand that is a common hurdle for people to overcome when getting started with this on Ubuntu, but the last thing I expect when running docker containers is for it to make configuration changes on my host.

I'm all for making things easier for people to get started, but this version of ` docker_run.sh` goes way beyond just a docker-run command.

aviadra:
I must agree with you.
I myself have neglected this way, because it seems to not do what it
should, and other stuff mentioned.

PromoFaux:
Closing as stale - Feel free to reopen if this needs revisting

PR #683, opened 2020-09-28
Make buster the default debian version

lightswitch05:
Make buster the default debian version. Stretch is LTS and the tag still exists for those who need it, but its time to move on (IMHO).

## Description
Make the `latest` and non-os-specific tags use the buster version. Keep the stretch tags around for those who need it. This will probably break things for some people who are relying on the stretch version but are not using the stretch tags, the OS-specific tags are still kinda new and never got a lot of attention - so I doubt many people are using it. I've marked it as a 'breaking' change because of this, but isn't exactly breaking since people can just update their tags to use whichever version they like.

* `latest` -> `buster`
* `vX.X.X` -> `buster`
* `vX.X.X-buster` -> `buster`
* `vX.X.X-stretch` -> `stretch`

## Motivation and Context

Stretch is LTS, Buster is current, Bullseye is around the corner.

## How Has This Been Tested?

I've been running buster for a couple months now without issue, but I haven't tested this tag change specifically. Hopefully Github Actions will yell at me if I missed something.

## Types of changes

## Checklist:


dschaper:
I would like to see this happen. I think it's time for the change.

dschaper (approved):

dschaper:
@diginc When ever you're ready.

diginc (approved):

Issue #720, opened 2020-12-03
[MegaThread] Settings not persisting between updates / docker-compose down then up etc


PromoFaux:
Going to merge a few threads and pin this one, to hopefully prevent duplicate reports:

It is something we are aware of - the advice I can give at the moment is to use [environment variables](https://github.com/pi-hole/docker-pi-hole#environment-variables) in your `docker run`/`docker-compose.yml` (or however one manages their containers.

Obviously that doesn't cover cases such as saving more than 2 upstream DNS servers, and a few other things. Personally I agree that there could be some persistence in user-changes to `setupVars.conf` and/or from the web interface. but digging through the comments on all the below issues I found this comment by @diginc : in #643 

> I have a [rewrite for config management in the works](https://github.com/pi-hole/docker-pi-hole/blob/beta-v5.0-refactor/docker/bash_functions.sh#L18-L24) (that I stalled out on) for better allowing both of these use cases to exist simultaneously but it's a pretty big breaking change to the environment variable schema and requires lots of people change their ENVs and re-writing/testing alot of the startup code.

And he makes a very good point.

In the mean time: Environment variables are your friend

## Existing threads:
- No Save Settings | No Files #198
- Feature Request: support DNS3 and DNS4 items for IPv6 upstream DNS servers #262
- Environment vars forcing use changes to reset #283
- Pi-hole settings lost at docker reboot #441
- PiHole Docker doenst use correct DNS Servers (always defaults to googles) #505
- Chosen upstream DNS servers not saved #604
- Changes to WEBPASSWORD are ignored #643
- Settings not being persistet across restarts #668
- Some DNS settings being lost on docker restart #687
- Serious issue with setupVars.conf file #719

PromoFaux:
Comment from @xBellaDonna in #718 

> I can confirm I also have this issue, affecting more than just conditional forwarding. I first noticed this when watchtower pulled a new image and restarted my container, disabling both DNSSEC and conditional forwarding, and preventing me from accessing several local services by hostname.
> 
[...]

PR #721, opened 2020-12-04
Prevent unset env vars overwriting setupvars.conf values

PromoFaux:
## Description
See discussion in #720 ([here](https://github.com/pi-hole/docker-pi-hole/issues/720#issuecomment-738685000)

## Motivation and Context
People expect to be able to set settings via the web interface, and have them preserved between container builds.

While the obvious answer may be to expect people to just set environment variables in their compose/run scripts - this may not be obvious to some. 

This is not an entirely complete solution - as it misses things like DNS1 and DNS2 (and even the ability for a user to have, say, 16 upstream DNS servers selected. I've no idea why they would, but you can in core so you should be able to with docker, I guess!)

## How Has This Been Tested?
Not as a container - but the theory seems sound

## Types of changes

## Checklist:

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -55,29 +55,32 @@ load_web_password_secret
 generate_password
 validate_env || exit 1
 prepare_configs
-change_setting "PIHOLE_INTERFACE" "$PIHOLE_INTERFACE"
-change_setting "IPV4_ADDRESS" "$IPV4_ADDRESS"
-change_setting "QUERY_LOGGING" "$QUERY_LOGGING"
-change_setting "INSTALL_WEB_SERVER" "$INSTALL_WEB_SERVER"
-change_setting "INSTALL_WEB_INTERFACE" "$INSTALL_WEB_INTERFACE"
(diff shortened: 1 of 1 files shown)


xBelladonna:
Since the check for empty strings in my implementation isn't used in this version, the parameter expansion braces are no longer required, but it makes no functional difference, only cleaner code. In any case I've tested this fix with both the repo unit tests and in a container running real world tests and can confirm it works as intended.

PromoFaux:
@diginc would appreciate your thoughts/suggestions on this please

diginc:
Step in the right direction, though if someone sets a env variable and restarts it still overwrites variables which I know some people report as a 'bug'.

diginc (approved):

PromoFaux:
> if someone sets a env variable and restarts it still overwrites variables

Good angle, I hadn't considered that. So you're saying, for example, I initially set the container up using env variables, and then make a change in the web interface - the changed setting will be reverted back to the env variable on restart.

To be honest, if one is setting their config via env variables, they should not be surprised when their config is set to those env variables, but then... users to be users.. :D

PR #727, opened 2020-12-16
Re-jig how Pi-hole's upstream DNS servers are set

PromoFaux:
## Description

This PR is another change towards solving #720. It's often mentioned by users that PIHOLE_DNS_1 and _2 are reset on restarts and on rebuilds, falling back to google defaults.

I aim to solve this by introducing a new variable: `PIHOLE_DNS_`, which is defined as a list of IP addresses, delimited by a semicolon (e.g `8.8.8.8;8.8.4.4` - custom ports also accepted, e.g `127.0.0.1#5353`)

A benefit of this over using the existing `DNS1` and `DNS2` variables is that users are not limited to using just 2. If they wish, they can set as many as they like. It also means that they do not have to set `DNS2` to `no` when only wanting to use 1 upstream server.

If `PIHOLE_DNS_` is not set, then any existing `DNS1` and `DNS2` variables will be converted to the `PIHOLE_DNS_` variable. If `DNS1` and `DNS2` are not set, then the script will fall back to the defaults of `8.8.8.8` and `8.8.4.4`

## How Has This Been Tested?
Built the container locally (what seems like thousands of times whilst I debugged 😉) and ran various combinations of the above

## Types of changes

(not breaking as previous variables are converted)

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,8 +102,7 @@ There are other environment variables if you want to customize various things in
 | `ADMIN_EMAIL: <email address>`<br/> *Optional Default: ''* | Set an administrative contact address for the Block Page
 | `TZ: <Timezone>`<br/> **Recommended** *Default: UTC* | Set your [timezone](https://en.wikipedia.org/wiki/List_of_tz_database_time_zones) to make sure logs rotate at local midnight instead of at UTC midnight.
 | `WEBPASSWORD: <Admin password>`<br/> **Recommended** *Default: random* | http://pi.hole/admin password. Run `docker logs pihole \| grep random` to find your random pass.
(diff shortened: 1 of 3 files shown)


PromoFaux:
Some gifs:

No env vars set, nothing in `setupVars.conf` (fallback to default):
defaultdns

`DNS1` set to `1.1.1.1` and `DNS2` set to `no`:
dns1and2

`PIHOLE_DNS_` set to `3.3.3.3;4.4.4.4;127.0.0.1#1234`:
dnsnewvar

No env vars set, existing values detected in `setupVars.conf`
dnsalreadyinsetupvars

diginc:
Looks good, nice testing screenshots.  Lets get some more testers using this on `:dev` tag next

diginc (approved):

Issue #486, opened 2019-08-01
FR: DNS-over-TLS


eku:
This is a... 

- [x] Request for a new or modified feature
- [ ] Issue trying to run the docker image
- [ ] Issue trying to build / test / develop the docker image

## Description

Support for DNS-over-TLS in the default docker image would be great. But please with the possibility to use your own DNS server and not 1.1.1.1 because of privacy concerns.

https://docs.pi-hole.net/guides/dns-over-https/

https://github.com/getdnsapi/stubby

heartbleeded:
Did you manage to setup dockerized pihole and DNS-over-TLS?

Joostvanderlaan:
I found these 2 projects trying to accomplish this: 
https://github.com/eldridgea/erised
https://github.com/kquinsland/skyhole

mcarson75:
I have modified the Pi-Hole docker image to also include unbound, which you can configure to use DNS-over-TLS with whichever servers you want:

[https://github.com/mcarson75/unbound-pihole](https://github.com/mcarson75/unbound-pihole)

Docker image mcarson75/unbound-pihole

CumpsD:
Would be nice to have this in the `pihole/pihole` image

pladaria:
To me this is a must because my ISP is intercepting DNS calls

buckaroogeek:
I found a simple DNS-Over-HTTPS client written in Go along with a companion docker compose yaml file and rolled that into my existing docker compose yaml used to manage a pi-hole instance, among other services. Documented at https://github.com/buckaroogeek/unifi-pihole-doh-docker.

jol64:
I don´t really want to hijack that issue, but I think it is pretty close to what I want.
Actually I tried to combine docker images for pihole and stubby (I used https://hub.docker.com/r/mvance/stubby, but looking at the complexity I´d opt for https://hub.docker.com/r/flexo3001/stubby). The issue I ran into is that I cannot reference the stubby container as DNS1: stubby#8053 (or #53 depending on what the container exposes) because dnsmasq expects an IP address rather than a DNS name. If you could convert the name to IP address within https://github.com/pi-hole/docker-pi-hole/blob/master/bash_functions.sh, then I assume it would be kind of trivial to get DNS-over-TLS.
Thanks, Joachim

b13sch:
Hi, please add DoT in the default docker image! I'm interested, too! How can we push this?

dschaper:
[...]

Issue #425, opened 2019-02-18
[Feature Request]: Use Docker container names as upstream DNS-Sevrers instead of IP only


Ichnafi:
This is a... 

- [x] Request for a new or modified feature
- [ ] Issue trying to run the docker image
- [ ] Issue trying to build / test / develop the docker image

## Description
You can set manually two custom DNS server IPs in the GUI. This have to be valid IPv4/IPv6 adresses.
If you use a DNS proxy (i.e. Stubby) in another docker container as upstream DNS server for Pi-Hole, you can not enter the container name into those fields and use the docker internal DNS, so pi-hole can forward request to the DNS-proxy. Hence you have to give the DNS-proxy container a static IP-Address, which is not an optimal use of the docker infrastructure.

## Expected Behavior
One can use either a valid IPv4/IPv6 address as an upstream server, or a docker container name making use of docker's internal DNS to resolve the upstream dns's IP in the docker virtual network.

## Actual Behavior
One can only enter valid IPv4/IPv6  into the GUI as custom upstream servers. If you don't there is a error message

## Steps to Reproduce and debugging done

e.g. your docker run command, pages to visit, CLI commands you ran
1.
2. 
3. 
4. 

## Debug steps I have tried

- [ ] I have tried destroying my container instance, pulling the newest image version, and re-creating a new container
- [ ] I have tried running the nearly stock `docker run` example in the readme (removing any customizations I added)
- [ ] I have tried running without my volume data mounts to eliminate volumes as the cause
- [ ] I have searched this repository for existing issues and pull requests that look similar 

## Your Environment 

* Docker Host Operating System and OS Version: Raspbian Stretch
* Docker Version: 18.09.2
* Hardware architecture: ARMv7

yegle:
This is especially useful for people running pihole in docker. Otherwise people have to manually configure network and allocate IP addresses.

As a workaround here's what I'm doing:

First create a network to put all your containers that requires access to pihole:
```
$ docker network create --subnet 172.29.0.0/16 infra
```
Then use the following docker compose file:
```
version: '2'
services:
  pihole:
    container_name: pihole
    image: pihole/pihole:latest
    restart: unless-stopped
    ports:
      - 53:53/tcp
      - 53:53/udp
      - 67:67/udp
    environment:
      - TZ=America/Los_Angeles
      - DNS1=172.29.1.2  # this is the stubby container IP address
      - DNS2=no
[...]

PR #729, opened 2020-12-17
modification to allow for container names (without ports) instead of …

jol64:
this pull request allows to use container names as DNS1 and DNS2 instead of IP addresses. It thus addresses https://github.com/pi-hole/docker-pi-hole/issues/486 and https://github.com/pi-hole/docker-pi-hole/issues/425.
The change does not support name#port, but in a docker environment this is imho less of a need. ip#port is still supported (untested). However it is likely incompatible with https://github.com/pi-hole/docker-pi-hole/pull/727.

## Types of changes

## Checklist:

not sure about style..
Feedback welcome.


PromoFaux:
Thanks for the PR, a couple things to mention : 

Pulls should be based on, and made against the `dev` branch

I also think this could be made to be compatible with #727 (I may be biased as the author of that pull but I would like to see us move in that direction to give more flexibility

Perhaps, [this](https://github.com/pi-hole/docker-pi-hole/pull/727/files#diff-65a5db1ae8f8e01ebdbbfafcf4e24618ad7cffe036bba92dce1390383e5c73e5R89-R90) could be adjusted to read (something like)

```bash
    [ -n "${DNS1}" ] && echo "Converting DNS1 to PIHOLE_DNS_" && PIHOLE_DNS_="$(dig +short ${DNS1} || ${DNS1})"
    [[ -n "${DNS2}" && "${DNS2}" != "no" ]] && echo "Converting DNS2 to PIHOLE_DNS_" && PIHOLE_DNS_="$PIHOLE_DNS_;$(dig +short ${DNS2} || ${DNS2})"
```

Though that's totally untested and would likely need tweaking.

diginc:
Introducing this new DNS#X variable would break lot of people using current ENV functionality I think.  Ideally we don't have a new variable at all and we just check the contents of the current DNS# variables to see if it's IP or hostname and behave appropriately (lookup or not).  #727 is moving forward into dev (which I updated your MR to target) so you might need to pull dev and rework this a little based off the below feedback too.

Testing from my ubuntu host dig wasn't returning false / failing to backup value on an IP address in my testing so content checking the result of dig might be required.

```
$ dig +short dns.google.com
8.8.8.8
8.8.4.4
$ dig +short 8.8.8.8 || echo test  # (no output)
$ dig +short 8.8.8.8 ; echo return value $?
return value 0
$ dig +short 8.8.8.8#5353 ; echo return value $?
return value 0
```

Another complication is multiple DNS return values as shown in the above example.  This would need to be advertised as 'container DNS support' rather than full DNS support until multi-line DNS values being returned was handled.  Stating the limitation in docs is fine for a first pass but having full DNS support eventually would be cool to avoid any confusion from users.

jol64:
> Introducing this new DNS#X variable would break lot of people using current ENV functionality I think.

A local variable? Ok I am neither a shell expert nor did I use a complex environment.

> Testing from my ubuntu host dig wasn't returning false / failing to backup value on an IP address in my testing so content checking the result of dig might be required.

I only did some quick tests. Actually to me it was more like a proof of concept than productization.
 
> Another complication is multiple DNS return values as shown in the above example. This would need to be advertised as 'container DNS support' rather than full DNS support until multi-line DNS values being returned was handled. Stating the limitation in docs is fine for a first pass but having full DNS support eventually would be cool to avoid any confusion from users.

Totally fine, and for that purpose actually I assume others could just benefit without this pull really being merged. My current docker container for pi-hole just uses the standard image and replaces the script file (also a reason why I forked master rather than dev - I wanted it to be stable).

Imho the real question is whether @PromoFaux  might want to pick up the idea if that is going to be merged? I don´t see a value in perfection of something that is going to be replaced likely..

diginc:
> A local variable? Ok I am neither a shell expert nor did I use a complex environment.

my bad, you're right that's fine. Still hadn't had my coffee.

> Imho the real question is whether @PromoFaux might want to pick up the idea if that is going to be merged? I don´t see a value in perfection of something that is going to be replaced likely..

Your approach of checking dig result + using IP if there's no hostname result approach should still work - [possibly between 99 and 100 here?](https://github.com/pi-hole/docker-pi-hole/pull/727/files#diff-65a5db1ae8f8e01ebdbbfafcf4e24618ad7cffe036bba92dce1390383e5c73e5R99).  

Something like..

```
lookup_i="$(dig +short $i)"
test -n "$(lookup_i)" && i="$($lookup_i | head -1)"  # first value only
```

Getting modifications merged upstream helps the overall community and prevents you from running into future roadblocks when the mainline code strays far enough from a forked customized volume mount that something breaks.

PromoFaux:
Superseded by https://github.com/pi-hole/docker-pi-hole/pull/727

PR #736, opened 2020-12-22
remove no longer required code

PromoFaux:
Removes code no longer needed as the fix was merged into core/master a while back

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -89,12 +89,6 @@ else
     fetch_release_metadata "${webInterfaceDir}" "${WEB_VERSION}"
 fi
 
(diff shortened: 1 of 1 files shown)


PromoFaux:
@diginc, have tested `:dev`, all appears to be working. Should be ready for merge to master then tag now.

lightswitch05 (approved):

DL6ER:
I should maybe mention that `armel` isn't something we have with FTL. The referenced commit has long been replaced:

https://github.com/pi-hole/pi-hole/blob/master/automated install/basic-install.sh#L2391-L2479

I don't see any docker-specific code.

PR #743, opened 2020-12-27
Add `hostname` to docker-compose

e2jk:
## Description
Add `hostname` to docker-compose
Note that in docker-compose-jwilder-proxy.yml example file, I used the same FQDN `pihole.yourDomain.lan` instead of just `hostname`, as used in that file for the DEFAULT_HOST environment variable (not sure if `.yourDomain.lan` should be included).

## Motivation and Context
If this is not set, the PiHole admin page shows the Docker container's ID as hostname instead.
Suggestion originally found [here](https://www.reddit.com/r/pihole/comments/gagn7p/container_hostname_for_dockerpihole/fozwt5a/). This PR addresses the next comment there ("That directive isn't in the Github notes (see https://github.com/pi-hole/docker-pi-hole)")

## How Has This Been Tested?
Tested on my own installation of PiHole, installed from the example Docker Compose file with no specific customizations.

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ services:
   pihole:
     container_name: pihole
     image: pihole/pihole:latest
-    hostname: pihole
+    hostname: pi.hole
     ports:
       - "53:53/tcp"
       - "53:53/udp"
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ services:
   pihole:
     container_name: pihole
     image: pihole/pihole:latest
+    hostname: pihole
     ports:
       - "53:53/tcp"
       - "53:53/udp"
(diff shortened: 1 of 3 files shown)


e2jk:
I've changed the proposed hostname from `pihole` to `pi.hole` to be consistent with the parameter which actually was already defined in the [docker_run.sh](https://github.com/pi-hole/docker-pi-hole/blob/2bd15dea8feff6743924cc9898aebc2f7b9f4c15/docker_run.sh#L19) file.
Still leaving the jwilder-proxy example to pihole.yourDomain.lan to be consistent with the rest of that file's content, namely DEFAULT_HOST.

PromoFaux (approved):

Issue #595, opened 2020-04-03
Historical Data: PHP Fatal error: Allowed memory size exhausted


lue30499:
**In raising this issue, I confirm the following:**

- [x] I have read and understood the [contributors guide](https://github.com/pi-hole/AdminLTE/blob/master/CONTRIBUTING.md).
- [X] The issue I am reporting can be *replicated*.
- [X] The issue I am reporting isn't a duplicate (see [FAQs](https://github.com/pi-hole/pi-hole/wiki/FAQs), [closed issues](https://github.com/pi-hole/AdminLTE/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20), and [open issues](https://github.com/pi-hole/AdminLTE/issues)).

**How familiar are you with the the source code relevant to this issue?:**

2

---
**Expected behaviour:**

All of my >800k historical data to show

**Actual behaviour:**

Does show short term data but data larger than a 4 weeks worth errors with:

In web browser debug log
Server Error 500
and in "/var/log/lighttpd/error.log"
"FastCGI-stderr: PHP Fatal error: Allowed memory size of 134217728 bytes exhausted"

**Steps to reproduce:**

In pihole 4.x it would show all my data. Ill I did was upgrade to the beta. (I am running this in docker)

I did try a fresh container and copied the pihole-FTL.db into it with the same results.

**Debug token provided by [uploading `pihole -d` log](https://discourse.pi-hole.net/t/the-pihole-command-with-examples/738#debug):**

1rxn14fgtj

**Troubleshooting undertaken, and/or other relevant information:**
I am running the beta.

All that happened was an upgrade. I did fix the error and make all my data show by putting:
" ini_set('memory_limit', '1024M'); "
At the top of the " /var/www/html/admin/api_db.php "

ayasystems:
I have the same issue

2020-05-07 10:42:00: (mod_fastcgi.c.2543) FastCGI-stderr: PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 606208 bytes) in /var/www/html/admin/api.php on line 139

dschaper:
The default memory allocation for PHP is 128M, you'll need to increase that limit if you want to access larger time periods.

ayasystems:
Where should be changed?? REgards!

lue30499:
@ayasystems 

I was able to fix the issue by hoping into the container (docker exec -it <docker ID> bash)

[...]

Issue #592, opened 2020-03-25
Expose ability in docker container to specify php memory


evanrich:
Running the latest container here, on kubernetes.  I've given the pod a request of 256MB( raised to 1G), and a limit of 4G of ram, yet when the container runs, and I click on certain things like "long query" or "show all" in query log, I get an error "An unknown Error has occured". Looking at lighttpd error log, I see this

```
 Allowed memory size of 134217728 bytes exhausted (tried to allocate 16777224 bytes) in /var/www/html/admin/api_db.php on line 172
2020-03-24 19:09:42: (mod_fastcgi.c.2543) FastCGI-stderr: PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 16777224 bytes) in /var/www/html/admin/api_db.php on line 172
```

Can you expose a variable to set for PHP memory usage?

dannytsang:
I would also like this feature. I have also run into errors running queries.

Something like exposing a variable which can be set on a startup script prior to starting the webserver.

dschaper:
Inside the container, at `/etc/php/7.0/cgi/php.ini` is `memory_limit = 128M` where you'd make the change. For now you might be able to volume map that file to something better for your environment.

I'm not sure if setting that value to `-1` in the image would work, that tells PHP to use all available. If you use that as the value then you could set the container's available memory and see if that's stable.

BenMatheja:
I was also just wondering how to go on from here. Editing it inside the container will work for some time but will get overridden at next pull

stephdl:
I hit myself this bug too, when I do a `Long term data > Query log`  more than the week, then I have this message in lighthttpd log

`FastCGI-stderr: PHP Fatal error: Allowed memory size of 134217728 bytes exhausted`

php.ini of cli is set to unlimited (-1), cgi php.ini is set to 128MB, it is not enough, either add a env Var or set to unlimited

stephdl:
same issue here : https://github.com/pi-hole/docker-pi-hole/issues/595

PR #745, opened 2020-12-28
Add env memory_limit  to php cgi

stephdl:
## Description
I figured that I cannot browse the history log (Long term data > Query log) when I set more than one week I got an error in lighthttpd log, these issues are related

https://github.com/pi-hole/docker-pi-hole/issues/592
https://github.com/pi-hole/docker-pi-hole/issues/595

the error : 

`FastCGI-stderr: PHP Fatal error: Allowed memory size of 134217728 bytes exhausted`

php.ini of cli is set to unlimited (-1), cgi php.ini is set to 128MB, it is not enough, either add a env Var or set it to unlimited

## Motivation and Context

Since when we pull a new image I will erase my manual change I would like to use a ENV var to modify on the fly the cgi php limit

## How Has This Been Tested?

Well new to docker and more in pihole project, I would be more than happy to get some guidance to build and test locally

## Types of changes

## Checklist:


bash_functions.sh
```diff
@@ -181,6 +181,12 @@ setup_php_env() {
 
     echo "Added ENV to php:"
     grep -E '(VIRTUAL_HOST|ServerIP|PHP_ERROR_LOG)' "$PHP_ENV_CONFIG"
+
+    if [[ -n ${PHP_ENV_CGI_MEMORY_LIMIT} ]]; then 
+        grep -qP 'memory_limit = 128M' || \
```
Haarolean, line 186: What if it changes here sometimes? Wouldn't it be better to place an additional config file under php directory with `.ini` extension with desired settings? Like proposed [here](https://github.com/pi-hole/docker-pi-hole/issues/595#issuecomment-699701827)

stephdl: Nice hack, It will probably fix the issue for sure. Thank to bring it however I tend to think that a bug should not be fixed by the enduser with a config file mounted by a custom volume

I would hope to get an Env var to modify at the container creation.

I am involved in another community project (NethServer) and this kind of modification are done with a simply esmith property. We try to provide a solution for the most common usages without issues for the system administrator

stephdl:
Just tested it, indeed it works as expected, create a configuration file and mount it as a volume. I found that the reliable value to set is hard to determine because I tested `512M` and It was not enough, I have to set it to `1024M`.

Thank for the input

stephdl:
Thank for your inputs, lets close

Issue #490, opened 2019-08-11
Lighttpd does not start in container


haioken:
This is a... 

- [ ] Request for a new or modified feature
- [x] Issue trying to run the docker image
- [ ] Issue trying to build / test / develop the docker image

## Description

When starting the container, container gets stuck attempting to start `lighttpd -D -f /etc/lighttpd/lighttpd.conf` over and over again.

## Expected Behavior

lighttpd should start and web server becomes available

## Actual Behavior

Logs from issue (docker logs pihole): https://pastebin.com/Vu069ZdH

## Steps to Reproduce and debugging done

e.g. your docker run command, pages to visit, CLI commands you ran
1. Clean any existing container with `docker rm pihole`
2. Clean any existing image with `docker rmi pihole/pihole`
3. Run container with `docker run -d --name pihole -p 53:53/tcp -p 53:53/udp -p 81:80 -e TZ="America/Chicago" -v "$(pwd)/etc-pihole/:/etc/pihole/" -v "$(pwd)/etc-dnsmasq.d/:/etc/dnsmasq.d/" --dns=127.0.0.1 --dns=1.1.1.1 --restart=unless-stopped pihole/pihole:latest`
4. Attempt to start lighttpd manually (same result) with `docker exec -it pihole lighttpd -D -f /etc/lighttpd/lighttpd.con`

## Debug steps I have tried

- [x] I have tried destroying my container instance, pulling the newest image version, and re-creating a new container
- [x] I have tried running the nearly stock `docker run` example in the readme (removing any customizations I added)
- [x] I have tried running without my volume data mounts to eliminate volumes as the cause
- [x] I have searched this repository for existing issues and pull requests that look similar, none found regarding this message

I have tried entering the container and running apt-get update/apt-get upgrade, tried enabling/disabling lighttpd modules.

## Context and extra information
Had run this container previously on centos host, but host has been reinstalled with openSUSE tumbleweed. Other containers I run seem unaffected.

## Your Environment 

* Docker Host Operating System and OS Version: openSUSE Tumbleweed 20190806
* Docker Version: 19.03.1, build 74b1e89e8ac6
* Hardware architecture: amd64

kekec777:
Did anybody resolve this ?

iamskok:
@haioken have you resolved this issue?

areese:
I hit this with Podman, it turned out the my issue was missing the ```--dns 127.0.0.1``` before any other dns, but that is in the bug report.

[...]

PR #748, opened 2021-01-02
Add DNS to sample docker-compose.yml to reflect readme instructions

jeremyrempel:
## Description
Update the sample `docker-compose.yml` to reflect the instructions in readme

## Motivation and Context
- Failure to correctly add the DNS can result in issues that are difficult to debug
- Others have encountered the same issue see #490

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,9 @@ services:
       - "67:67/udp"
       - "80:80/tcp"
       - "443:443/tcp"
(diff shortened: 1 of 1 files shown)


mew1033:
Pretty sure this is actually wrong now as per: #762.

PromoFaux:
This is no longer required.

PR #757, opened 2021-01-15
Prepare for new release

PromoFaux:
bump to core 5.3.2 / web 5.3


install.sh
```diff
@@ -4,7 +4,7 @@ mkdir -p /etc/pihole/
 mkdir -p /var/run/pihole
 # Production tags with valid web footers
 export CORE_VERSION="$(cat /etc/docker-pi-hole-version)"
-export WEB_VERSION="${CORE_VERSION}"
+export WEB_VERSION="v5.3"
 export PIHOLE_SKIP_OS_CHECK=true
 # Overwrite WEB_VERSION if core and web versions are different
 #export WEB_VERSION="v5.1.1"
```
dschaper, line 10: I think this is where the version diff is supposed to go, though I don't quite see why it's that way.

PromoFaux: oh. I see, yeah I guess it could go there too. I didn't see that before. I'll redo

PR #767, opened 2021-01-20
Add SYS_NICE capability so FTLDNS can increase its priority

cyounkins:
## Description

See https://github.com/pi-hole/FTL/pull/798

Sets the capability in `docker-compose.yml` files and updates the documentation.

## Motivation and Context

Without this capability, users will see `WARNING: Required Linux capability CAP_SYS_NICE not available` in FTL log

## How Has This Been Tested?

I run a `docker-compose.yml` file with `SYS_NICE`. FTLDNS then sets the correct priority and the warning message is avoided.

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ services:
     #   https://github.com/pi-hole/docker-pi-hole#note-on-capabilities
     cap_add:
       - NET_ADMIN
+      - SYS_NICE
     restart: unless-stopped
 ```
 2. Run `docker-compose up --detach` to build and start pi-hole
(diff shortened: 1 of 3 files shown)


cyounkins:
I need to modify `fix_capabilities` in `bash_functions.sh` to handle the case when FTLDNS is run as non-root.

I see a general issue here. When running FTLDNS as non-root, `setcap CAP_NET_ADMIN` would fail if it isn't explicitly given to the container, making it not recommended as described in the documentation but required when running as non-root. NET_RAW would not be able to be explicitly dropped.

I can see a few ways to handle the capabilities, including SYS_NICE:
1. Try to `setcap` in `fix_capabilities` and fail silently. Later in execution IF those caps are needed, an error (ex for NET_BIND_SERVICE) or warning (SYS_NICE) will occur.
2. Require `setcap` for all possibly-needed capabilities, allowing the user to somehow specify an override if they choose to not provide it (ex SYS_NICE) or explicitly drop it (NET_RAW)
3. Mix and match the above. Fail hard when lacking NET_BIND_SERVICE because I don't think there's a use case without it, and warn or be silent if failing for others.

dschaper:
I don't think this will actually do anything in reality.

Setting SYS_NICE is just going to adjust the niceness inside the container, correct? And that's not doing much since it's the only real process running inside the container. 

For the NET_ADMIN, that's the only way to get `pihole-FTL` to bind to port 53 if run non-root.

cyounkins:
Interesting! Testing confirms you are correct - the process niceness is only relative to other processes within the container. I must say I'm surprised since `top` on the host shows the different priorities/niceness of the processes and the kernel is shared in the docker environment.

For the curious, here's how I tested: https://gist.github.com/cyounkins/633f5bec6623d8d466412f405617c083

dschaper:
The processes do show on the host but I'm pretty sure the cgroup inside the namespace limits what the container can do.

PR #774, opened 2021-01-26
Remove PIHOLE_DNS_* from setupVars if DNS* == "no"

### LINKED ISSUE #773: Docker "-e DNS2: no" doesn't disable the secondary upstream DNS
### Versions
  Pi-hole version is v5.2.4 (Latest: v5.2.4)
  AdminLTE version is v5.3.1 (Latest: v5.3.1)
  FTL version is v5.5.1 (Latest: v5.5.1)

### Platform
- OS and version:  ArchLinux
- Platform:  Docker

### Expected behavior

Documentation on [hub.docker.com](https://hub.docker.com/r/pihole/pihole) says regarding env variable `DNS2`
```
DNS2: <IP>Optional Default: 8.8.4.4 | Secondary upstream DNS provider, default is google DNS, no if only one DNS should used
```

### Actual behavior / bug

If I set `DNS2: no` I still get it enabled under Settings-DNS (Google 8.8.4.4) and it's **being used as Primary DNS** until I go to Settings-DNS and click "Save" only after that queries are being forwarded to my custom DNS firstly.

### Steps to reproduce

docker-compose.yml
```
  pihole:
    image: pihole/pihole:latest
    ports:
      - '53:53/tcp'
      - '53:53/udp'
      - '67:67/udp'
    cap_add:
      - NET_ADMIN
    dns:
      - 127.0.0.1
      - 192.168.1.1
    environment:
      DNS1: 192.168.1.1
      DNS2: "no"
    restart: always
```

### Additional context
Content of /etc/pihole/setupVars.conf
```
IPV6_ADDRESS=0:0:0:0:0:0
PIHOLE_DNS_2=8.8.4.4
LIGHTTPD_ENABLED=true
CACHE_SIZE=10000
INSTALL_WEB_SERVER=true
INSTALL_WEB_INTERFACE=true
IPV4_ADDRESS=0.0.0.0
PIHOLE_DNS_1=192.168.1.1
WEBPASSWORD=<removed>
PIHOLE_INTERFACE=eth0
QUERY_LOGGING=true
BLOCKING_ENABLED=true
```

sprnza:
Resolves #773 

For detail see the linked issue.

## Checklist:

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -90,17 +90,23 @@ if [ -z "${PIHOLE_DNS_}" ]; then
     # a semi-colon delimited string and store in PIHOLE_DNS_
     # They are not used anywhere if PIHOLE_DNS_ is set already
     [ -n "${DNS1}" ] && echo "Converting DNS1 to PIHOLE_DNS_" && PIHOLE_DNS_="$DNS1"
-    [[ -n "${DNS2}" && "${DNS2}" != "no" ]] && echo "Converting DNS2 to PIHOLE_DNS_" && PIHOLE_DNS_="$PIHOLE_DNS_;$DNS2"
+    [ -n "${DNS2}" ] && echo "Converting DNS2 to PIHOLE_DNS_" && PIHOLE_DNS_="$PIHOLE_DNS_;$DNS2"
 fi
 
-# Parse the PIHOLE_DNS variable, if it exists, and apply upstream servers to Pi-hole config
+# Parse the PIHOLE_DNS variable, if it exists, and apply upstream servers to Pi-
+hole config
 if [ -n "${PIHOLE_DNS_}" ]; then
     echo "Setting DNS servers based on PIHOLE_DNS_ variable"
     # Split into an array (delimited by ;)
     PIHOLE_DNS_ARR=(${PIHOLE_DNS_//;/ })
     count=1
     for i in "${PIHOLE_DNS_ARR[@]}"; do
-        change_setting "PIHOLE_DNS_$count" "$i"
+        if [[ "$i" == "no" ]] ; then
+            delete_setting "PIHOLE_DNS_$count"
+            unset "PIHOLE_DNS_$count"
+        else
+            change_setting "PIHOLE_DNS_$count" "$i"
+        fi
         ((count=count+1))
     done
 else


start.sh
```diff
@@ -90,17 +90,23 @@ if [ -z "${PIHOLE_DNS_}" ]; then
     # a semi-colon delimited string and store in PIHOLE_DNS_
     # They are not used anywhere if PIHOLE_DNS_ is set already
     [ -n "${DNS1}" ] && echo "Converting DNS1 to PIHOLE_DNS_" && PIHOLE_DNS_="$DNS1"
-    [[ -n "${DNS2}" && "${DNS2}" != "no" ]] && echo "Converting DNS2 to PIHOLE_DNS_" && PIHOLE_DNS_="$PIHOLE_DNS_;$DNS2"
+    [ -n "${DNS2}" ] && echo "Converting DNS2 to PIHOLE_DNS_" && PIHOLE_DNS_="$PIHOLE_DNS_;$DNS2"
 fi
 
-# Parse the PIHOLE_DNS variable, if it exists, and apply upstream servers to Pi-hole config
+# Parse the PIHOLE_DNS variable, if it exists, and apply upstream servers to Pi-
```
PromoFaux, line 96:
```suggestion
# Parse the PIHOLE_DNS variable, if it exists, and apply upstream servers to Pi-hole config
```

start.sh
```diff
@@ -90,17 +90,23 @@ if [ -z "${PIHOLE_DNS_}" ]; then
...
+hole config
```
PromoFaux, line 97:
```suggestion
```

PromoFaux (changes requested):

PromoFaux:
Thanks for the pull. I'll need to look it over properly later. Also to note, that all Pulls should be made against the `dev` branch.

I'm sure I tested the eventuality of `DNS2=no` when I deprecated those vars. The idea being that it shouldn't even make it into the `PIHOLE_DNS_` array to add it.

sprnza:
> The idea being that it shouldn't even make it into the
Then 8.8.4.4 is being added as the second (or primary?) DNS

PromoFaux:
OK, so what is actually happening is that any existing `PIHOLE_DNS_*` values in `setupVars.conf` (if it is persisted) are persisted between container instances. 

I tested this by setting the `PIHOLE_DNS_` env var to `1.1.1.1;2.2.2.2;3.3.3.3;4.4.4.4;5.5.5.5;6.6.6.6;7.7.7.7;8.8.8.8`, starting the container, then setting it just to `1.1.1.1;2.2.2.2`, and the rebuilding the container. The `3.3.3.3`+ values remained in place.

So the question really, is, do we want to delete _all_ other values in `setupVars.conf` if there are only two in the environment variables of the compose file/run script? 

I'll close this for now, but tagging in @diginc  and @dschaper  for further discussion

dschaper:
We'd need to decide who is the source of truth. 

I'd do something like:

if the environment variable is set then ignore setupVars.conf version of the variables.

PromoFaux:
so, basically: (pseudocode)
```
if (PIHOLE_DNS_ env var is set)
{
  delete PIHOLE_DNS_* from setupVars.conf
  set based on PIHOLE_DNS_ env var
}
else
{
  leave everything as we find it
}
```

dschaper:
I wouldn't delete anything.

```
if ENVVAR then
  create PIHOLE_DNS_ array from envvar
else
  create PIHOLE_DNS_ array from setupVars.conf
fi

use PIHOLE_DNS_ array
```

User can (and should) be able to use the web interface to modify DNS servers. But changes to DNS servers should not persist IF they start the new container with explicitly declared ENVVARs. Always use what they declare on the `docker run` command line or yaml Environment: lines.

If they don't want to declare explicitly, then use the setupVars.conf version and let them modify that as they wish.

dschaper:
Or the bash way:

>${parameter:-word}
>
>    If parameter is unset or null, the expansion of word is substituted. Otherwise, the value of parameter is substituted.
>
>${parameter:=word}
>
>    If parameter is unset or null, the expansion of word is assigned to parameter. The value of parameter is then substituted. Positional parameters and special parameters may not be assigned to in this way.

PR #806, opened 2021-02-26
Remove port 443 from templates and documentation.

dschaper:
## Description

We don't use 443 inside the container.

Fixes #755 

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #808, opened 2021-02-26
Remove port 443 from templates and documentation. (Redo)

PromoFaux:
See #806 for detail


sixcorners:
@PromoFaux It looks like a reference to 443 was missed. The docker port documentation in the README.md says "Ports to expose (53, 80, 67, 443), the bare minimum ports required for Pi-holes HTTP and DNS services"

PromoFaux:
Ah yes, so it was. Cheers for the heads up

PR #816, opened 2021-03-25
Start: Support hostnames and domains in PIHOLE_DNS_

networkException:
## Description
Made the start script resolve any hostnames in PIHOLE_DNS_ enviornment variable the changing setting the pihole setting. The readme has also been updated to include a hostname example.

## Motivation and Context
I noticed being unable to use pihole in a docker-compose stack network together with a dns proxy as the PIHOLE_DNS_ didn't respect hostnames.

## How Has This Been Tested?
I tested the changes locally by bind mouting my updated start.sh into a pihole container to allow more rapid testing. Any domain and hostname got resolved and custom ports specified with `#port`got respected as well.

## Types of changes

## Checklist:


kylekurz:
@PromoFaux I think this PR might fix the other side of my issue post update. I'm building a merge of your dev branch and this PR to test.

start.sh
```diff
@@ -100,7 +100,18 @@ if [ -n "${PIHOLE_DNS_}" ]; then
     PIHOLE_DNS_ARR=(${PIHOLE_DNS_//;/ })
     count=1
     for i in "${PIHOLE_DNS_ARR[@]}"; do
-        change_setting "PIHOLE_DNS_$count" "$i"
+        # Split upstream address by #
+        split=(${i//#/ })
+
+        # Resolve hostname if any, otherwise use ip
+        resolved="$(ping -q -c 1 -t 1 "$split" | grep PING | sed -e "s/).*//" | sed -e "s/.*(//")"
```
PromoFaux, line 107: I like this idea, but I'm not sure this is the right way to go about it.

Could we use `dig +short "$split"`, instead? Less pipes is always a good thing!
pseudocode:

```
resolved="$(dig +short "$split")"

if (resolved is blank or empty)
    resolved=split 
fi

### continue as below
```

PromoFaux (changes requested):

kylekurz (changes requested):
This works, if you remove any quotes around the `PIHOLE_DNS_` line, but it's breaking from the old way, so it either needs very clearly defined in the README, or adjusted to pull off the any string quoting characters.

Works: `PIHOLE_DNS_=cloudflared#5054`
Broken: `PIHOLE_DNS_='cloudflared#5054'`

I think we need to adjust before merging.

PromoFaux:
> but it's breaking from the old way,

I wasn't aware quote were needed in the current way?

image

kylekurz:
> > but it's breaking from the old way,
> 
> I wasn't aware quote were needed in the current way?
> 
> image

They're not NEEDED, but they're supported. So if you were quoting before, you *must* change after this gets merged, it breaks even IP address configurations.

PromoFaux:
Ah right, I get you now. It's a wonder any of my systems work at all - I'm basically blind when it comes to docker stuff ;)

start.sh
```diff
@@ -100,7 +100,18 @@ if [ -n "${PIHOLE_DNS_}" ]; then
     PIHOLE_DNS_ARR=(${PIHOLE_DNS_//;/ })
     count=1
     for i in "${PIHOLE_DNS_ARR[@]}"; do
-        change_setting "PIHOLE_DNS_$count" "$i"
+        # Split upstream address by #
+        split=(${i//#/ })
```
kylekurz, line 104: Per my earlier comment, this also needs to make sure to strip out string delimiters, otherwise you'll likely break existing installs, even if they're not using hostnames.

start.sh
kylekurz, line 107: Should we also filter to only do this if non-digits are found (e.g. this is actually a hostname)? Otherwise we're wasting time and networking to ping/dig something we don't need to. With ping, in particular, there's now a requirement that the system accepts pings and is up at launch, which wasn't a requirement before to start up.

kylekurz (changes requested):

start.sh
PromoFaux: I think this function is available in `start.sh`, as it [sources `webpage.sh`](https://github.com/pi-hole/docker-pi-hole/blob/master/start.sh#L44), which in turn [sources `basic-install.sh`](https://github.com/pi-hole/pi-hole/blob/master/advanced/Scripts/webpage.sh#L27)

https://github.com/pi-hole/pi-hole/blob/master/automated%20install/basic-install.sh#L1020-L1038

start.sh
dschaper: Can you give me an example of the input and the expected output for these?

I think there's some bash that is getting in the way, it does unexpected things when you start double quoting and single quoting and mixing in brace expansions and the like.

kylekurz: I've not tried double-quotes on the old image, but my existing compose file had:

```
PIHOLE_DNS_='127.0.0.1#5053'
```

And this code tried to ping `'127.0.0.1`. So anyone using string delimiters would have to adjust. I'd rather see this strip those at the beginning/end of the string.

dschaper: Okay, can you give me a list of what you want as acceptable input, and then exactly what you want that to be as output please?

networkException:
I will look into all the suggested changes. In my private pihole instance I just had some issues regarding the feature I added here however, the ping command reported unable to resolve :/

I'm not sure what could have caused this, I never experienced the issue before

start.sh
kylekurz: @dschaper looking at the bash script in its entirety, I think I'm just lucky that my quoted environment var worked. If the delimiters get down to this level from the Docker environment vars, there's nothing that I see that would have stripped them and my luck was that I only had one item, not a list. That said, I still think this feels like breaking behavior, even if it's unrelated to this PR and more just something my testing exposed. Could be a separate issue to open and fix?

Here's my input suggestions:
a) `127.0.0.1`
b) `'127.0.0.1'`
c) `"127.0.0.1"`
d) `127.0.0.1;1.1.1.1`
e) `'127.0.0.1;1.1.1.1'`
f) `"127.0.0.1;1.1.1.1"`
g) `cloudflare#5054`
h) `'cloudflare#5054'`
i) `"cloudflare#5054"`
j) `cloudflare#5054;1.1.1.1`
k) `'cloudflare#5054;1.1.1.1'`
l) `"cloudflare#5054;1.1.1.1"`

Expected output array for a, b, c: 
```
127.0.0.1
```

Expected output array for d, e, f:
```
127.0.0.1
1.1.1.1
```

Expected output array for g, h, i:
```
cloudflare#5054
```

Expected output array for j, k, l
```
cloudflare#5054
1.1.1.1
```

dschaper: Awesome! Thanks for the breakdown. 

I personally don't see the utility in having a FQDN as the DNS entry and it seems counterintuitive, but if we can do this in a way that doesn't add a whole lot of complexity then I don't want to block it. 

Let me play around a bit and see what shakes loose. I would like to try to make this all POSIX shell as well. Improving portability and removeing the need for `bash`ism's is a goal I'd like to reach.

kylekurz: @dschaper this is for docker swarm. If you have a container service named `cloudflare`, resolving inside the container will give you the right private network IP to hit that container.

dschaper: Doesn't swarm have it's own DNS in Docker Engine?

kylekurz: Probably, I'm new to swarm. If you can show me how to specify the right address for the *other* service when I spin up my pihole service, I'll gladly do that. But I've verified that the following don't work:

1. Putting in `127.0.0.1#5054` (I'm running a single node, but without network_mode: host, this just hits the pihole container)
2. Putting in `<host_ip>#5054` (Gets a response, but from an unexpected IP, which is the docker private network IP of the cloudflare container)

When I drop into a shell in the pihole container though, I can `dig +short cloudflare` and get the right IP, so bringing this in allows specifying the service name in the DNS entry and then passing the right IP to pihole every time we launch.

dschaper: Have you tried `dig +short cloudflare @127.0.0.11` to see what comes back? 127.0.0.11 is the IP for the Engine DNS.

kylekurz: I'm not sure how that helps me set up the yaml file. Resolving from the running container is "easy", but I need a way to force the container to do that. Allowing the processing of the PIHOLE_DNS_ variable to handle that is the only way I can see that working?

dschaper: What are you trying to achieve? 

(At the very least, `dig` is a lot easier to parse than `curl | sed | sed | sed`.

dschaper: Looking over the inputs listed I see things that would never be configured.

```
127.0.0.1
1.1.1.1
```

You wouldn't do that since Pi-hole would be using `localhost` and `1.1.1.1` as upstreams and you really can't control what gets resolved with each resolver. 

Same with:

```plain
cloudflare#5054
1.1.1.1
```

Why use `cloudflare` at the same time as `1.1.1.1`. You don't know what queries are going to go to what upstream.

kylekurz: For my uses, PiHole is the primary DNS, providing local DNS entries and the usual blocking. Instead of upstreaming to Google or Cloudflare over UDP, I upstream to a local instance of `visibilityspots/cloudflared`, which implements DoH for any external requests. So my swarm runs two containers, pihole and cloudflared, with this docker-compose.yml (edited to remove secrets):

```
version: "3.1"
services:
  pihole:
    hostname: pihole
    image: pihole/secure:latest
    depends_on:
      - cloudflared
    volumes:
      - /home/pi/docker-compose/pihole/docker:/etc/pihole
    environment:
      - ServerIP=192.168.25.233
      - TZ=PST
      - PIHOLE_DNS_=cloudflared#5054
    ports:
      - "53:53/tcp"
      - "53:53/udp"
      - "67:67/udp"
      - "80:80/tcp"

  cloudflared:
    image: visibilityspots/cloudflared:alpine-3.12
    ports:
      - "5054:5054/udp"
```

I don't care how the resolution happens in the start.sh script, just that it happens AFTER cloudflared is running, and I don't have a way to pin that IP or get it directly for the yaml file.

kylekurz: > Looking over the inputs listed I see things that would never be configured.
> 
> ```
> 127.0.0.1
> 1.1.1.1
> ```
> 
> You wouldn't do that since Pi-hole would be using `localhost` and `1.1.1.1` as upstreams and you really can't control what gets resolved with each resolver.
> 
> Same with:
> 
> ```
> cloudflare#5054
> 1.1.1.1
> ```
> 
> Why use `cloudflare` at the same time as `1.1.1.1`. You don't know what queries are going to go to what upstream.

Those were contrived examples to show that you may want to provide a list, either with/without port and with/without hostname. Just trying to fully cover the test cases, not providing real-world use cases in that. Would probably test with a 3-item list too, if I were to do it.

dschaper: Okay, and where does Swarm come in to play. I think you mentioned this is a single host so I'm not seeing why Docker Swarm is being used. 

And have you tried https://docs.docker.com/compose/compose-file/compose-file-v3/#depends_on to set Pi-hole to depend on cloudflared being up first?

kylekurz: I'm using docker secrets to manage the password, rather than specifying it in plaintext (See #824), so you can't use vanilla docker-compose and must use swarm, even if it's on a single node. Docker swarm doesn't support `network_mode: host`, so you have to actually provide the right IP for the upstream. My compose file listed above actually does already depend on `cloudflared` being up, but I don't know it's IP address when writing the file, only once it's booted.

dschaper: That's because:

> The `depends_on` option is ignored when deploying a stack in swarm mode with a version 3 Compose file.

And you can manage secrets in other ways that doesn't mean doing a single node swarm. 

I think this is a solution to a problem that is better handled in other ways and not by adding a breaking change to what we have now.

kylekurz: @dschaper this is why I said earlier I think this is probably a different issue, regardless of this PR. The fact is, it's already broken. If you set this in a compose file (pure docker-compose, no swarm involved):
`- PIHOLE_DNS_='1.1.1.1;1.0.0.1'`

You get this in /etc/pihole/setupVars.conf and the container will exit
```
PIHOLE_DNS_1='1.1.1.1
PIHOLE_DNS_2=1.0.0.1'
```

So whether we merge this or not, there's a problem with quotes, IMO. Once that's solved, I'd personally still like to see this PR come in for my use case, but I think we're conflating "add hostname support" with "fix quoting issues on environment variables".

dschaper: Okay, then we need to break this in to two different issues and PRs.

networkException: This uh got a lot more complicated then I though, I was happy to get hostname resolution working in bash with very limited knowledge on shell scripts anyways.

Could anyone else imagine to take over the branch themselves?

PromoFaux: Thank you, none the less, for rolling up your sleeves and coding your feature request - it's always nice to see people have a go at it themselves :)

kylekurz: @dschaper opened #838 to handle the cleaning side.

@networkException I'm happy to keep chiming in here, I've merged your branch to my fork, so if there's a decision to handle the other ticket, I'll finish cleanup on this PR (or, probably make a new one from my fork and replace this).

networkException: Thanks a lot everyone :^)

alpe12:
@networkException Did you find out how to make it work?
I've been breaking my head over this.

networkException:
Well I just use the patch I wrote for this PR in my setup (bindmount the start.sh) but that's not ideal

alpe12:
Maybe it would be possible for docker-compose to add a feature to translate host to ip? This way pihole would receive an IP.

Issue #824, opened 2021-04-11
Don't log password at boot if provided via Docker Secrets or environment


kylekurz:
If the docker container is started with either a WEBPASSWORD or WEBPASSWORD_FILE argument, don't print the password to logs. This defeats the use of standard secret protection practices. The only time the password should be printed to a log file is if the container generates it randomly at boot because the user failed to provide a password.

kylekurz:
While I have opened #834 after this merged, I was able to verify this was included in the merge, so I believe this can be closed.

PR #825, opened 2021-04-11
Issue #824 - Don't log password

kylekurz:
When setting the password, explicitly disable bash logging. Leave the
re-enable code so that other functions work as expected. Additionally,
do not remove the print in generate_password so randomly generated
passwords are still logged for user consistency.

## Description
I moved the `set +x 2>/dev/null` from below the password setup to above it to disable bash debug. I also removed the explicit echo of the password. The generate_password function still contains the expected echo line for the generated password value.

## Motivation and Context
Generally, passwords should not be logged. The only time that's appropriate is when the system is generating the password for the user.

## How Has This Been Tested?
I ran the included tests, as well as testing the resulting image in my development environment. 

## Types of changes
- [X ] Bug fix (non-breaking change which fixes an issue)

## Checklist:
- [X ] My code follows the code style of this project.

### DIFF
--- a/bash_functions.sh
+++ b/bash_functions.sh
@@ -230,17 +230,14 @@ setup_web_password() {
     setup_var_exists "WEBPASSWORD" && return
 
     PASS="$1"
-    # Turn bash debug on while setting up password (to print it)
+    # Explicitly turn off bash printing when working with secrets
+    { set +x; } 2>/dev/null
+
     if [[ "$PASS" == "" ]] ; then
         echo "" | pihole -a -p
     else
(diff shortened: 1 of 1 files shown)


PromoFaux:
So I've got this clear:

`Assigning random password: [password]` will still display, right? Kinda important for the user to be able to get in on first build 😏

kylekurz:
@PromoFaux yes! I left that in place, I just removed the logging when the actual `pihole -a -p` commands are made. Random password generation still logs the generated password as expected.

PR #828, opened 2021-04-13
Feature: add debian bullseye

lightswitch05:
Adding the new 'bullseye' Debian flavor - which is now in 'Hard Freeze' status. This isn't an important change, but every time there is a new Debian release, there is an influx of people wanting the newest version. If we have something already available, then it might be useful for testing before it eventually becomes the default version.

## Description

Just added Bullseye.

## Motivation and Context

Be available for testing by people who care about those things.

## How Has This Been Tested?

The first test is to see if Github Actions can build it or not.

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/test-and-build.yaml
+++ b/.github/workflows/test-and-build.yaml
@@ -21,7 +21,7 @@ jobs:
       fail-fast: false
       matrix:
         ARCH: [amd64, armhf, arm64]
-        DEBIAN_VERSION: [stretch, buster]
+        DEBIAN_VERSION: [stretch, buster, bullseye]
     env:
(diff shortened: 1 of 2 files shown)


richlander:
Is it possible the PR title is wrong?

PromoFaux:
Doesn't look like anything to me... 🙂

richlander:
For the project I work on (.NET), we've already switched to bullseye for our latest in-development release. It is working great. https://github.com/dotnet/dotnet-docker/blob/main/src/runtime-deps/6.0/bullseye-slim/amd64/Dockerfile

Issue #834, opened 2021-04-15
pihole/pihole:v5.8 won't start


kylekurz:
This is a: Bug

## Details

In an attempt to vet the PR you merged for me earlier today, I pulled in the latest pihole/pihole image and attempted to run on two different machines, both failed with the following error:

`  [✗] DNS service is NOT listening
kill: usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill -l [sigspec]`

## Related Issues
- [ x] I have searched this repository/Pi-hole forums for existing issues and pull requests that look similar 

## How to reproduce the issue 

1. Environment data
  * Operating System: Raspbian
  * Hardware: Pi 4B and Pi 3B+ 
  * Kernel Architecture: 
  * Docker Install Info and version: 
    - Software source: Official docker-ce `Docker version 20.10.5, build 55c4c88` 
    - Supplimentary Software: `docker-compose version 1.29.0, build unknown` 
  * Hardware architecture: `Linux pihole 5.10.17-v7l+ #1403 SMP Mon Feb 22 11:33:35 GMT 2021 armv7l GNU/Linux` 

2. docker-compose.yml contents, docker run shell command, or paste a screenshot of any UI based configuration of containers here:
Here's my compose file that works if I go back to version 5.2.4 that I had from before the upgrade:
```
version: "3"
services:
        pihole:
                container_name: pihole
                domainname: docker
                hostname: pihole
                image: pihole/pihole:v5.8
                network_mode: host
                dns:
                        - 127.0.0.1
                        - 1.1.1.1
                restart: unless-stopped
                volumes:
                        - /home/pi/docker-compose/pihole/docker/pihole/pihole:/etc/pihole
                cap_add:
                        - NET_ADMIN
                environment:
                        - ServerIP=192.168.25.5
                        - TZ=PST
                        - WEBPASSWORD=changeme
                        - DNS1='127.0.0.1#5053'
                        - DNS2=''

        cloudflared:
                container_name: cloudflared
                domainname: docker
                image: visibilityspots/cloudflared:alpine-3.12
                network_mode: host
                restart: unless-stopped
                environment:
                        - PORT=5053
```
3. any additional info to help reproduce
I have a more complex setup involving ansible and docker swarm/stack that I'm attempting to deploy, but I backed down to this to try to remove my issues as much as possible and limit it to the new container.

## These common fixes didn't work for my issue

- [ x] I have tried removing/destroying my container, and re-creating a new container
[...]

PR #836, opened 2021-04-16
Move gravity on boot to a @reboot cron job

mdzw:
## Description
This is a proposed solution to #834.

## Motivation and Context
Gravity update is relying on pihole-FTL for its DNS resolution. If "gravity on boot" is enabled, we shouldn't try to run it before the long-running instance of pihole-FTL is starting.

Copied from my comment on the issue:

I believe the correct way to solve this issue is to change when "Gravity on Boot" is run.

The s6 init system has [different stages](https://github.com/just-containers/s6-overlay#init-stages). Currently, "gravity on boot" is [run](https://github.com/pi-hole/docker-pi-hole/blob/master/s6/debian-root/etc/cont-init.d/20-start.sh#L18) during Stage 2.ii: **cont-init.d**. One instance of pihole-FTL is started during cont-init, but it is only there to check the validity of the config files; it exits soon after starting. The final "service mode" instance of pihole-FTL is not started until **Stage 2.iii**, when the supervisor starts doing work.

If gravity.sh is counting on FTL to do its DNS lookups, then we should not run gravity until the supervised instance of FTL is running. We can accomplish this by **moving "Gravity on Boot" to a `@reboot` line in /etc/cron.d/gravity-on-boot**, and making that file's existence dependent on the value of SKIPGRAVITYONBOOT.

This will work because cron isn't started until we've reached Stage 2.iii.

edit: PR incoming.

## How Has This Been Tested?
A thorough examination of the startup sequence for Docker pihole, gravity, and pihole-FTL.

## Types of changes

## Checklist:

- [x] My code follows the code style of this project. (Maybe?)


mdzw:
It looks like the cron job is not getting run (there's no `/var/log/pihole_updateGravity.log` file), but I don't see any indication of why.

PromoFaux:
> there's no `/var/log/pihole_updateGravity.log` file

Sure about that?

https://github.com/pi-hole/docker-pi-hole/pull/836#issuecomment-821268702

PromoFaux:
What _does_ happen, though, is that setting `SKIPGRAVITYONBOOT` causes it to boot loop:

image

mdzw:
> What _does_ happen, though, is that setting `SKIPGRAVITYONBOOT` causes it to boot loop:

d'oh, probably should check for that file before trying to `rm` it, or put `|| true` at the end.

PromoFaux:
Initially, though, it does appear to fix this issue. I can't make it fall over with this build in the same way I was able to make it fall over yesterday with v5.8.

Can you sort the issue with `SKIPGRAVITYONBOOT`, and then we can try this on the `:dev` tag to see if it fixes it for others

PromoFaux:
Yep, looks to be working, thanks. Now we wait for the CI to finish.

PromoFaux (approved):
Go for `:dev` we can always revert if needed

PR #843, opened 2021-04-19
`install.sh` overhaul

PromoFaux:
## Description
Big ol' refactor of `install.sh` basically mimicking the way I personally test fresh installs of varying branches/tags:
- Clone repos locally
- Checkout branch/versioned tag
- Run install script as --unattended --reconfigure
- ???
- profit

## Motivation and Context

I started off looking at how to get a nightly build going based on `development`/`devel` branches of the main repos, and ended up doing this instead (I still plan on doing that eventually, but it relies on this)

## How Has This Been Tested?

Built locally as I went along

Example screenshot showing the versions defined in `VERSIONS` being checked out at the end of it all.
image

## Types of changes

## Checklist:

Edit: Also shaves 3 mins off total CI time...


lightswitch05 (approved):
I can't say I followed the logic 100% though, but everything looks fine to me. `VERSIONS` vs. `VERSION` is kinda confusing. Could we just ditch `VERSION` in favor of the new and improved `VERSIONS`? Or does that add too much complexity to this change set?

VERSIONS
```diff
@@ -0,0 +1,3 @@
+CORE_VERSION=v5.3.1
+WEB_VERSION=v5.5
+FTL_VERSION=v5.8
```
DL6ER, line 3:
```suggestion
FTL_VERSION=v5.8.1
```

PromoFaux:
>Could we just ditch VERSION in favor of the new and improved VERSIONS?

Yes and no. It's currently tangled up in the image naming - which I didn't want to touch just yet (though it probably should go, I just haven't worked out how to untangle it yet)

install.sh
```diff
@@ -47,53 +66,29 @@ PH_TEST=true . "${PIHOLE_INSTALL}"
...
-fi
-
 sed -i 's/readonly //g' /opt/pihole/webpage.sh
 sed -i '/^WEBPASSWORD/d' /etc/pihole/setupVars.conf
 
 # Replace the call to `updatePiholeFunc` in arg parse with new `unsupportedFunc`
 sed -i $'s/helpFunc() {/unsupportedFunc() {\\\n  echo "Function not supported in Docker images"\\\n  exit 0\\\n}\\\n\\\nhelpFunc() {/g' /usr/local/bin/pihole
 sed -i $'s/)\s*updatePiholeFunc/) unsupportedFunc/g' /usr/local/bin/pihole
+sed -i $'s/)\s*piholeCheckoutFunc/) unsupportedFunc/g' /usr/local/bin/pihole
+sed -i $'s/)\s*piholeCheckoutFunc/) unsupportedFunc/g' /usr/local/bin/pihole
```
PromoFaux, line 91: Duplicate

```suggestion

```

lightswitch05:
I wouldn't mind taking a stab at reducing the number of version files. I don't think it will be so complicated, just a matter of finding the time to do it

PromoFaux:
Please do! From my experiments it is only used when the images are built for testing - IIRC, it doesn't actually make it out into the real world, as the tag is used for the image names on docker hub

PromoFaux:
> reducing the number of version files.

Thinking about this.. we might even not need the new `VERSIONS` file, they're only used in `install.sh`, so we may as well just declare them at the top of that... What do you think?

lightswitch05:
I like the new versions file and think it would be really nice to have all the versions declared in a single place. I'll try to do the updates required- but it will take a couple weeks before I have the time to look at it. In the mean time, having two is acceptable

PR #861, opened 2021-05-29
Single versions file

lightswitch05:
## Description
Update tests and build to only require the new `VERSIONS` file. I added python library `python-dotenv` to handle parsing the file.

## Motivation and Context
No need to have a `VERSION` file and a `VERSIONS` file. Just doing some cleanup. Remove the old `VERSION` file.

## How Has This Been Tested?
Ran the tests and build. As far as I know these files have nothing to do with runtime.

## Types of changes

- [x] Tech debt cleanup

## Checklist:


Dockerfile.py
```diff
@@ -20,12 +20,16 @@
...
-    raw_version = v.read().strip()
-    __version__ = raw_version.replace('release/', 'release-')
+CORE_VERSION = None
+
+
+def read_pihole_versions():
+    global CORE_VERSION
+    dot = os.path.abspath('.')
+    config = dotenv_values('{}/VERSIONS'.format(dot))
+    CORE_VERSION = config['CORE_VERSION']
```
PromoFaux, line 32: Just a small niggle... totally up for discussion. As we base the containers tag version on whatever the Pi-hole FTL version is, I wonder - should we pull in the FTL version here instead of Core?

lightswitch05: I don't really have an opinion one way or the other. I've never fully understood the versioning of this project- or even pihole as a whole. My goal here was to match the existing logic using the new file, but if you think it's an opportunity for an improvement, then that's fine with me too.

PromoFaux: Generally the only time the version number is the same across the three main components is at milestone releases, which get a major version bump. From there it's [semantic versioning](https://semver.org/) as each project receives patches/minor features etc.

Docker's tag versions used to follow whatever version the core repo was at, though in recent releases that was bumped to be equivalent to FTL instead, as a) it's generally at a higher version number than core, b) it's fast become the most central part of the project

So yeah, maybe an opportunity for a bit of consistency?

lightswitch05: I updated it @PromoFaux - thanks for the clarification! I thought I had matched existing functionality, but I was clearly wrong. Please let me know if any further enhancements are needed.

PromoFaux: > I thought I had matched existing functionality, but I was clearly wrong. 

Absolutely not wrong at all :) The tagged version number of the image has been based on FTL version number, but the version number in the `VERSION` file has always been the core repo version number.  Convolution is fun ;)

PromoFaux (approved):
LGTM 

image

PR #870, opened 2021-06-27
add cloudflared for doh in pihole docker setup

frankiexyz:
add cloudflared for doh in pihole docker setup

## Description
I was trying to add the cloudflared as a DOH lookup into the existing container

## Motivation and Context
pihole docker container is using plaintext DNS at the moment which is not ideal, user can change the pihole DNS server to 127.0.0.1#5533 if they would like to use DOH

## How Has This Been Tested?
- docker build in x86_64, pi 3b 
```
'sudo docker exec -ti pi-hole-cloudflared dig @127.0.0.1 -p5533 archlinux.org'

; <<>> DiG 9.11.5-P4-5.1+deb10u5-Debian <<>> @127.0.0.1 -p5533 archlinux.org
; (1 server found)
;; global options: +cmd
;; Got answer:
;; ->>HEADER<<- opcode: QUERY, status: NOERROR, id: 220
;; flags: qr rd ra; QUERY: 1, ANSWER: 1, AUTHORITY: 0, ADDITIONAL: 1

;; OPT PSEUDOSECTION:
; EDNS: version: 0, flags:; udp: 4096
;; QUESTION SECTION:
;archlinux.org.                 IN      A

;; ANSWER SECTION:
archlinux.org.          3600    IN      A       95.217.163.246

;; Query time: 202 msec
;; SERVER: 127.0.0.1#5533(127.0.0.1)
;; WHEN: Sun Jun 27 07:16:12 CDT 2021
;; MSG SIZE  rcvd: 71
```

## Types of changes

## Checklist:

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -30,13 +30,25 @@ fetch_release_metadata() {
 }
 
 apt-get update
-apt-get install --no-install-recommends -y curl procps ca-certificates git
+apt-get install --no-install-recommends -y curl procps ca-certificates git gnupg2
+
 # curl in armhf-buster's image has SSL issues. Running c_rehash fixes it.
 # https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=923479
 c_rehash
 ln -s `which echo` /usr/local/bin/whiptail
 curl -L -s $S6OVERLAY_RELEASE | tar xvzf - -C /
 mv /init /s6-init
+
+# Download cloudflared
+export ARCH_TYPE="$(uname -m)"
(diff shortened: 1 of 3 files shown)


PromoFaux:
Thank you for the pull request, but this is going to be a no from me, I'm afraid. 

A better solution would be to run `cloudflared` in another container, rather than packaging it into the Pi-hole container. I do this at home with `unbound`.

DL6ER:
I agree, docker works by doing only one thing in a container. DoH is naturally separated from Pi-hole as it is "just" yet another upstream DNS provider.

Issue #874, opened 2021-07-05
Container failing to start correctly with "kill usage" message


deosrc:
This is a: Bug

## Details

The Pi-Hole container does not start correctly with the following "kill usage" message appearing in the logs (surrounding messages included for context):

```text
 ::: Docker start setup complete
kill: usage: kill [-s sigspec | -n signum | -sigspec] pid | jobspec ... or kill -l [sigspec]
  Pi-hole version is v5.3.1 (Latest: v5.3.1)
  AdminLTE version is v5.5 (Latest: v5.5)
  FTL version is v5.8.1 (Latest: v5.8.1)
[cont-init.d] 20-start.sh: exited 0.
```

## Related Issues
- [x] I have searched this repository/Pi-hole forums for existing issues and pull requests that look similar 

#555 
#405 

The above issues seem to be general support threads with various generic fixes. I am including specific details here on my findings so figured a new issue would be best. If maintainers believe this would be best in one of the above threads instead, let me know and I'll move the content.

## How to reproduce the issue 

1. Environment data
  * Operating System: RaspberryPi OS 
  * Hardware: RasPi4B
  * Kernel Architecture: ArmV7
  * Docker Install Info and version: 
    - Software source: K3s
    - Supplimentary Software: Kubernetes Dashboard
  * Hardware architecture: ARMv7

2. docker-compose.yml contents, docker run shell command, or paste a screenshot of any UI based configuration of containers here

```yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: pihole
  namespace: home
  labels:
    app: pihole
spec:
  replicas: 1
  selector:
    matchLabels:
      app: pihole
  serviceName: pihole
  template:
    metadata:
      labels:
        app: pihole
    spec:
      restartPolicy: Always
      hostNetwork: true
      dnsPolicy: "None"
      dnsConfig:
        nameservers:
          - 127.0.0.1
          - 1.1.1.1
      containers:
        - name: pihole
          image: pihole/pihole:v5.8.1
          securityContext:
            capabilities:
              add:
                - NET_ADMIN
          ports:
            - containerPort: 31180
              name: http
            - containerPort: 53
              protocol: UDP
              name: dns-udp
            - containerPort: 53
              protocol: TCP
              name: dns-tcp
            - containerPort: 67
              protocol: UDP
              name: dhcp
          volumeMounts:
            - name: config
              mountPath: /etc/pihole/
            - name: dnsmasq
[...]

PR #875, opened 2021-07-05
Resolve TODOs around killing pihole-FTL

deosrc:
## Description

Resolving TODO comments around preventing `kill -9`.

## Motivation and Context

Log output showing kill usage is causing confusion in debugging issues (see #874).

## How Has This Been Tested?

1. Tested as a separate script with only echo statements and changing process to known running/not running.
2. Executed instructions in TESTING.md to build container.
3. Ran container from step 2. Checked for log message output and that web interface is available and showing values.

## Types of changes

Tech Debt

## Checklist:


PromoFaux:
Thanks for the submission - and that certainly makes things tidier / more versbose (which is a good thing when it comes to the container startup)

But a note to add that the `TODO` was actually around the usage of `kill -9` altogether. The contention being that it's rather like using a sledgehammer to crack a nut...

I'll check this PR fully later after work and merge if it doesn't break anything... 😏

deosrc:
> But a note to add that the `TODO` was actually around the usage of `kill -9` altogether. The contention being that it's rather like using a sledgehammer to crack a nut...

Makes sense. I've re-added the todo comment specifically referring to the argument.

deosrc:
Going to close this as the main issue this was addressing was in the startup which looks to have been re-written to better resolve this.

PR #876, opened 2021-07-07
Enable Debian Bullseye (11) images

lightswitch05:
## Description

Enable Debian Bullseye (11) images. This PR is blocked on upstream changes, but I wanted to go ahead and get it queued up so it is available when ready.

* Actual support: https://github.com/pi-hole/pi-hole/pull/4221
* Test-level support: https://github.com/pi-hole/pi-hole/pull/4126

Once upstream is ready, I can rebase to re-run the actions before merge.

## Motivation and Context

Have Bullseye images available.

## How Has This Been Tested?

I haven't tested this since the changes are only in github actions and will be apparent on PR submission.

## Types of changes

## Checklist:

We could probably reference these images in the README if we really want to, but it might be prudent to hold off on that until we're more confident in our ability to support Bullseye. At some point far in the future, we'll make the Bullseye image the 'default' version.

### DIFF
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         ARCH: [amd64, armhf, arm64, i386]
-        DEBIAN_VERSION: [stretch, buster]
+        DEBIAN_VERSION: [stretch, buster, bullseye]
     env:
       ARCH: ${{matrix.ARCH}}
       DEBIAN_VERSION: ${{matrix.DEBIAN_VERSION}}
--- a/.github/workflows/nightly-build.yml
+++ b/.github/workflows/nightly-build.yml
@@ -11,6 +11,8 @@ jobs:
       matrix:
         ARCH: [amd64, armhf, arm64, i386]
         DEBIAN_VERSION: [stretch, buster, bullseye]
+        exclude:
+          - {ARCH: arm64, DEBIAN_VERSION: bullseye}
     env:
       ARCH: ${{matrix.ARCH}}
       DEBIAN_VERSION: ${{matrix.DEBIAN_VERSION}}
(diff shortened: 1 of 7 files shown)


PromoFaux:
(unsolicited) Tip for testing `bullseye` locally:  

`VERSIONS`:
```
CORE_VERSION=development
WEB_VERSION=devel
FTL_VERSION=development
```

`DEBIAN_VERSION=bullseye ./gh-actions-test.sh`

image

Interesting thing to note from that, though, the `cat /etc/*release` from inside the container  does not have the version number in it. It doesn't matter so much here because we call `export PIHOLE_SKIP_OS_CHECK=true` in `install.sh`

lightswitch05:
Good info, however the latest bullseye image does have the release information set

PromoFaux:
Also, as a side note: You have write access to this repo AFAIK, so you should be able to just clone and push feature branches directly rather than via your fork. Makes it a tiny bit easier for one of us to test (like a tiny tiny bit)

PromoFaux:
Note for future: `bullseye` `arm64` is not building properly at the moment (not sure why, noticed that upstream had it excluded from the matrix, so have had to do the same.

Not sure if this will work, but you may need to add :

```yml
exclude:
  - {ARCH: amd64, DEBIAN_VERSION: bullseye}
```

to the `matrix:` section for the time being...

PromoFaux:
Oh wait, I can do it.. I have write access to your fork

PromoFaux:
Some more detail on arm64/bullseye: https://forum.armbian.com/topic/16740-debootstrap-base-system-second-stage-failed/

PromoFaux:
Closing as stale - Switching release image to bullseye should just be as simple as changing the value at the top of the root `Dockerfile` now. We don't need to provide images for different flavours of OS, IMHO. Sometimes I think about trying to make an alpine one... I don't think I'm bored enough yet!

jauderho:
So does this mean that the official images will be switching to bullseye?

Having an alpine based container would be interesting.

PromoFaux:
> So does this mean that the official images will be switching to bullseye? 

There is the potential for this, yes - but we will need to do some testing first I guess!

Issue #878, opened 2021-07-15
With a fresh container/volumes, dashboard is dead if SKIPGRAVITYONBOOT is set


DistractionRectangle:
This is a: Run Issue

## Details

The Web Interface shows a completely blank dashboard if gravity isn't initialized.
On bringing up a fresh container with empty volumes and `SKIPGRAVITYONBOOT` set, the dashboard fails to show any data. Upon updating gravity under `Tools -> Update Gravity` and navigating back to the dashboard, it becomes populated with data.

## Related Issues
- [X] I have searched this repository/Pi-hole forums for existing issues and pull requests that look similar 

I have not found an relevant issues in this repo, adminlte, main pihole repository or the forums; though I specifically searched for `SKIPGRAVITYONBOOT` as the keyword since it seems to be at the root of the issue.

## How to reproduce the issue 

1. Environment data
  * Operating System: **Ubuntu 20.04 LTS** 
  * Hardware: PC; AMD 3100, b550m itx
  * Kernel Architecture: x86/amd64
  * Docker Install Info and version: 
    - Software source: official docker-ce installed using their instructions
    - Supplimentary Software: NA
  * Hardware architecture: x86/amd64

2. docker-compose.yml contents, docker run shell command, or paste a screenshot of any UI based configuration of containers here

<details><summary><b>Erroneous docker-compose.yml</b></summary>
<p>

Command used to test docker-compose.yml changes
`docker-compose down && sudo rm -rf .docker-volumes/ && docker-compose up -d`

```YAML
services:
  pihole:
    cap_add:
    - NET_ADMIN
    environment:
      SKIPGRAVITYONBOOT: '1'
      ServerIP: 192.168.1.14
      WEBPASSWORD: password
    image: pihole/pihole:latest
    ports:
    - mode: host
      protocol: tcp
      published: 8080
      target: 80
    restart: unless-stopped
    volumes:
      - ./.docker-volumes/pihole/etc-dnsmasq.d:/etc/dnsmasq.d:rw
      - ./.docker-volumes/pihole/etc-pihole:/etc/pihole:rw
version: '3.7'
```

</p>
</details>

3. any additional info to help reproduce

## These common fixes ~~didn't~~ do work for my issue as detailed below. This is an edge case that can only be reached under specific circumstances, and more likely to be reached by new users since they'll have empty//uninitialized volumes.

- [X] I have tried removing/destroying my container, and re-creating a new container
[...]

PR #881, opened 2021-07-21
Ignore SKIPGRAVITYONBOOT if gravity.db does not exist

PromoFaux:
## Description
Adds an additional check on start. If `/etc/pihole/gravity.db` does not exist, then `SKIPGRAVITYONBOOT` will be ignored on this occasion so that the database may be generated.

## Motivation and Context

Fixes #878 

## How Has This Been Tested?

Built the container loads locally.

Fresh volume:

image

After `down && up`:

image

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

PR #883, opened 2021-07-23
Fix gravityDB check

DistractionRectangle:
## Description

Duplicate check performed in pi-hole/pi-hole/gravity.sh
pihole-FTL.conf exposes config to relocate gravityDB; it should
be sourced to determine where to check for gravityDBs existence

Signed-off-by: D.Rect <48034372+DistractionRectangle@users.noreply.github.com>

## Motivation and Context

Fixes: #878 
Follow up to: #881 

## How Has This Been Tested?

docker-compose up with empty volume,
00_gravitydb_not_initialized
docker-compose down and docker-compose up (persisting volume)
01_gravity_initialized

Using interactive bash shell in running container, move gravity.db with: 
`echo "GRAVITYDB=/etc/pihole/I_moved_gravity.db" > /etc/pihole/pihole-FTL.conf`

Followed by docker-compose down and docker-compose up (persisting volumes):
02_moved_gravity

And one more time to see if gravity check finds the db at the new location and skips gravity:
03_moved_gravity

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/s6/debian-root/etc/cont-init.d/20-start.sh
+++ b/s6/debian-root/etc/cont-init.d/20-start.sh
@@ -14,9 +14,19 @@ $bashCmd /start.sh
 if [ -n "$PYTEST" ]; then
     sed -i 's/^gravity_spinup$/#gravity_spinup # DISABLED FOR PYTEST/g' "$(which gravity.sh)"
 fi
-if [ -z "$SKIPGRAVITYONBOOT" ] || [ ! -f /etc/pihole/gravity.db ]; then
+
+gravityDBfile="/etc/pihole/gravity.db"
+config_file="/etc/pihole/pihole-FTL.conf"
+# make a point to mention which config file we're checking, as breadcrumb to revisit if/when pihole-FTL.conf is succeeded by TOML
+echo "  Checking if custom gravity.db is set in ${config_file}"
+if [[ -f "${config_file}" ]]; then
+    gravityDBfile="$(grep --color=never -Po "^GRAVITYDB=\K.*" "${config_file}" 2> /dev/null || echo "/etc/pihole/gravity.db")"
+fi
+
+
+if [ -z "$SKIPGRAVITYONBOOT" ] || [ ! -e "${gravityDBfile}" ]; then
     if [ -n "$SKIPGRAVITYONBOOT" ];then
-        echo "  SKIPGRAVITYONBOOT is set, however no gravity.db exists (Likely due to a fresh volume). This is a required file for Pi-hole to operate."
+        echo "  SKIPGRAVITYONBOOT is set, however ${gravityDBfile} does not exist (Likely due to a fresh volume). This is a required file for Pi-hole to operate."
         echo "  Ignoring SKIPGRAVITYONBOOT on this occaision."
     fi
 


s6/debian-root/etc/cont-init.d/20-start.sh
```diff
@@ -14,7 +14,22 @@ $bashCmd /start.sh
...
-if [ -z "$SKIPGRAVITYONBOOT" ] || [ ! -f /etc/pihole/gravity.db ]; then
+
+# gravity.sh consistent check for gravityDBfile's existence
+# GRAVITYDB may be overwritten by source pihole-FTL.conf below
+GRAVITYDB="/etc/pihole/gravity.db"
+
+# Source pihole-FTL from install script
+pihole_FTL="/etc/pihole/pihole-FTL.conf"
+if [[ -f "${pihole_FTL}" ]]; then
+  source "${pihole_FTL}"
```
DL6ER, line 25: Sourcing `pihole-FTL.conf` is bad because of mainly two things:

1. FTL allows for a multitude of comment styles, not only `#` at the beginning of a line.
  There are many users using `;` for comments because this is in line with that PHP's INI parsing expects. However, sourcing a file with such comments in it (with `root` privileges) can be very dangerous.
2. Pi-hole v6.0 will switch to [TOML](https://toml.io/) for the config file and `source` will not work as expected.

DistractionRectangle:
What would be the recommended way of checking the `GravityDB` setting in pihole-FTL.conf?
I'm following how it's normally checked in gravity.sh: [gravity.sh#L69](https://github.com/pi-hole/pi-hole/blob/4736e03108763cc2d5659f48d8a1e8a64d9b2608/gravity.sh#L69)

Does pihole -g/pihole updateGravity run at a different user level? Near as I can tell, docker-pi-hole currently touches that code path in gravity as root user currently: [20-start.sh#L18](https://github.com/pi-hole/docker-pi-hole/blob/76be304d79d8c8820575cf68128e0e6b53b57d82/s6/debian-root/etc/cont-init.d/20-start.sh#L18)

s6/debian-root/etc/cont-init.d/20-start.sh
DistractionRectangle: Regarding point 1.What would be the recommended way of checking the `GravityDB` setting in pihole-FTL.conf?
I'm following how it's normally checked in gravity.sh: [gravity.sh#L69](https://github.com/pi-hole/pi-hole/blob/4736e03108763cc2d5659f48d8a1e8a64d9b2608/gravity.sh#L69)

Does pihole -g/pihole updateGravity internally run as a different user? Near as I can tell, docker-pi-hole currently touches that code path in gravity as root user currently: [20-start.sh#L18](https://github.com/pi-hole/docker-pi-hole/blob/76be304d79d8c8820575cf68128e0e6b53b57d82/s6/debian-root/etc/cont-init.d/20-start.sh#L18)

Regarding point 2. In the original issue, I floated the idea of splitting out the initialization responsibilities from gravity.sh into it's own script mainly for this reason, to avoid maintaining the same functionality in both repos. 

In my head this would look like: refactoring gravity.db and local.list initialization into an `init_gravity.sh` located under `pi-hole/advanced/Scripts` and touching [`PI_HOLE_FILES` in `basic-install.sh`](https://github.com/pi-hole/pi-hole/blob/4736e03108763cc2d5659f48d8a1e8a64d9b2608/automated%20install/basic-install.sh#L73) to make sure [it's properly cleaned up](https://github.com/pi-hole/pi-hole/blob/4736e03108763cc2d5659f48d8a1e8a64d9b2608/automated%20install/basic-install.sh#L1427). Both gravity and docker could source the file that gets copied to the installation directory.

DL6ER: You could use something like
``` bash
GRAVITYDB="$(grep --color=never -Po "^GRAVITYDB=\K.*" "/etc/pihole/pihole-FTL.conf" 2> /dev/null || echo "/etc/pihole/gravity.db")"
```

which would also make the `if` unnecessary as the `|| echo` will account for a non-existing conf file. True, this would not (yet) account for a future TOML transition, but that's a problem of when this will happen. It's probably better to add something like
```
pihole-FTL which gravitydb
```
to get the answer independently from the particular language of the config file.

As for the other parts of the code: It's probably just done wrong there.

DistractionRectangle:
```
pihole-FTL which gravitydb
```

Errors out in my testing, running it in the container yields this output:
> root@78540f0a8f52:/# pihole-FTL which gravitydb
pihole-FTL: invalid option -- 'which'
Command: 'pihole-FTL which gravitydb'
Try 'pihole-FTL --help' for more information

Reading the last match off `/var/log/pihole-FTL.log` might be a way around the config file issue//eventual TOML transition, assuming it's stable enough to rely on:

```
gravityDBfile="$(grep --color=never -Po "GRAVITYDB: Using \K.*" "/var/log/pihole-FTL.log" 2> /dev/null | tail -1 || echo "/etc/pihole/gravity.db")"
```

DL6ER: > It's probably better **to add** something like
>
> ```
> pihole-FTL which gravitydb
> ```

Sorry, I see this was misleading - with *add* I meant we can add this to `pihole-FTL`. There is no such feature at the moment.

---

> Reading the last match off `/var/log/pihole-FTL.log` might be a way around the config file issue

I'm not familiar enough (not at all TBH) with this script is used. Is it guaranteed that FTL was started before (and had time to parse its config file)? Furthermore, is this script run only once (immediately at start) or can it be run also later on? In the latter case, `pihole-FTL.log` might have been rotated away and the information you are looking for is already gone.

PromoFaux: Kinda like:

image

-----

The script runs on container start,  but I wouldn't rely on grepping the log file for the location of `gravity.db`, I would just grep the config file...

DistractionRectangle: 😅 I just forced pushed the log based solution before reading this. Since FTL is started before and brought down after this check (originally because it was needed to run gravity before gravity was relegated to a cron job), it shouldn't be rotated away and should satisfy @DL6ER's concerns. 

I guess what I can do is output a warning if pihole-FTL.conf is not found, so that this isn't forgotten when it's replaced with TOML and grep the log file. I'll test and force push that shortly

DistractionRectangle:
Updated PR to reflect feedback. Instead of sourcing pihole-FTL.conf, it's grepped for a custom gravitydb entry. Some verbosity is added as a breadcrumb to make note of what config file we are grepping (as a reminder if/when it's succeeded by TOML). 

Testing proceed is the same as above:

With clean (empty) volumes:
`docker-compose up`
00_fresh_volume

`docker-compose down && docker-compose up`
01_gravity_exists

Running `echo "GRAVITYDB=/etc/pihole/I_moved_gravity.db" > /etc/pihole/pihole-FTL.conf` in a running container, followed by `docker-compose down && docker-compose up` on the host:
02_moved_gravity

And `docker-compose down && docker-compose up` one more time to see if found gravity at the new location:
03_check_moved

PromoFaux:
Thanks. I'm not 100% sure why the buster tests are now failing - as you haven't touched anything here. I'll try to take a look tomorrow, it's already late here!

s6/debian-root/etc/cont-init.d/20-start.sh
DL6ER: > I just forced pushed the log based solution before reading this

But this is not what is in your code. Is this correct?

DistractionRectangle: Correct, I pushed the log based solution just prior to reading PromoFaux's feedback:

> I wouldn't rely on grepping the log file for the location of gravity.db, I would just grep the config file...

And  updated again to reflect his preference to grep the config file; I initially didn't want to do that given it'll silently regress if/when it's moved to TOML, but added a line to the output to mention which config file is being grepped so the regression will be less silent if we do forget to update this in lockstep.

PromoFaux:
Sorry, have just started looking into the failed tests now. Seems to be an issue with the base images (which were updated about the same time the builds started failing)

Awaiting suggestions from the maintainer there https://github.com/multiarch/debian-debootstrap/issues/22

PromoFaux (approved):

PromoFaux:
Tests working again 😄

PR #885, opened 2021-07-27
Update README.md

gbrsni:
Small correction in the 'Quick Start' section of the README

## Description
No longer suggesting the usage of the '--detatched' flag as it does not work in certain versions of docker-compose (e.g. 1.17.1, the one included in the ubuntu repos).
Now suggesting the '-d' flag, which should work with every version.

## Motivation and Context
This is a very small change but I believe having a working 'Quick start' section is very important.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@ services:
       - NET_ADMIN
     restart: unless-stopped
 ```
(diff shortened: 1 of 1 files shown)


PromoFaux:
Hi there,

Thanks for the pull. Please remake this change based on, and against, the `dev` branch, Thanks

PR #897, opened 2021-08-14
Dependabot manage GHA versions.

dschaper:
## Description

Let Dependabot manage the GitHub Actions versions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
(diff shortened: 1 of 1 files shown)


jauderho:
Do you want Dependabot to scan for python dependencies too? In my forked repo, I noticed that urllib3 has an identified security issue.

See https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-33503

It also appears that the Pipfile has very specific pinned versions. It might be better to use ~= instead of == per https://pipenv-fork.readthedocs.io/en/latest/basics.html

I can submit additional pull requests if there is interest once this is merged.

PR #928, opened 2021-10-03
Use the namespace from secrets so others can build too.

dschaper:
## Description

Use DOCKERHUB_USER Instead of hard code `pihole/pihole` for the namespace. 

## Motivation and Context

Forks of this repo couldn't build as they would try to push to our namespace and fail.

## How Has This Been Tested?

Build it.

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/test-and-build.yaml
+++ b/.github/workflows/test-and-build.yaml
@@ -26,27 +26,34 @@ jobs:
     needs: test
     runs-on: ubuntu-latest
     steps:
-      - name: Checkout
+      -
+        name: Checkout
         uses: actions/checkout@v2
-      - name: Get the tag name
+      -
+        name: Get the tag name
         run: |
           echo "TAG=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
-      - name: Docker meta
+      -
+        name: Docker meta
         id: meta
         uses: docker/metadata-action@v3
         with:
           images: ${{ secrets.DOCKERHUB_NAMESPACE }}/pihole
           github-token: ${{ secrets.GITHUB_TOKEN }}
-      - name: Set up QEMU
+      -
+        name: Set up QEMU
         uses: docker/setup-qemu-action@v1
-      - name: Set up Docker Buildx
+      -
+        name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v1
-      - name: Login to DockerHub
+      -
+        name: Login to DockerHub
         uses: docker/login-action@v1
         with:
           username: ${{ secrets.DOCKERHUB_USER }}
           password: ${{ secrets.DOCKERHUB_PASS }}
-      - name: Build and push
+      -
+        name: Build and push
         uses: docker/build-push-action@v2
         with:
           context: .
--- a/.github/workflows/test-and-build.yaml
+++ b/.github/workflows/test-and-build.yaml
@@ -35,7 +35,7 @@ jobs:
         id: meta
         uses: docker/metadata-action@v3
         with:
-          images: ${{ secrets.DOCKERHUB_USER }}/pihole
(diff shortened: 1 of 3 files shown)


.github/workflows/test-and-build.yaml
```diff
@@ -26,34 +26,27 @@ jobs:
...
         run: |
           echo "TAG=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
-      -
-        name: Docker meta
+      - name: Docker meta
         id: meta
         uses: docker/metadata-action@v3
         with:
-          images: pihole/pihole
+          images: ${{ secrets.DOCKERHUB_USER }}/pihole
```
dschaper, line 38: May need to make this another secret in the case that namespace and user do not match.

dschaper: <html><body>


GITHUB_ACTOR | The name of the person or app that initiated the workflow. For example, octocat.
-- | --
GITHUB_REPOSITORY | The owner and repository name. For example, octocat/Hello-World.


</body>
</html>

PromoFaux: Possibly something [like this](https://github.com/pi-hole/FTL/blob/development/.github/workflows/build.yml#L50-L59)

Use like [this](https://github.com/pi-hole/FTL/blob/development/.github/workflows/build.yml#L92)

I think it is best to use custom secrets in the case of dockerhub. Using `github.repository`  in our case would be wrong, because here it is `pi-hole`, whereas on dockerhub it is `pihole`

PromoFaux:
``` yml
-       name: Determine Docker Hub Namespace
        id: namespace
        run: |
          DOCKERHUB_USER=${{ secrets.DOCKERHUB_USER }}
          DOCKERHUB_NAMESPACE=${{ secrets.DOCKERHUB_NAMESPACE }}          
          echo ::set-output name=DOCKERHUB_NAMESPACE::${DOCKERHUB_NAMESPACE:=${DOCKERHUB_USER}}

-       name: Docker meta
        id: meta
        uses: docker/metadata-action@v3
        with:
          images: ${{ steps.namespace.outputs.DOCKERHUB_NAMESPACE }}/pihole
```

etc etc

dschaper: That looks a bit hacky. Probably best to just have `secrets.DOCKERHUB_NAMESPACE`. Forks will still fail until that variable is set but at least it's configurable now.

PromoFaux: Yeah, indeed - if someone is going to need to add in `DOCKERHUB_USER` anyway, no point in having the check!

.github/workflows/test-and-build.yaml
```diff
@@ -26,34 +26,27 @@ jobs:
...
         run: |
           echo "TAG=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
-      -
-        name: Docker meta
+      - name: Docker meta
         id: meta
         uses: docker/metadata-action@v3
         with:
-          images: pihole/pihole
+          images: ${{ secrets.DOCKERHUB_NAMESPACE }}/pihole
```
PromoFaux, line 38: worth going one further and having `DOCKERHUB_REPO(NAME)`?, too?

dschaper: I don't know. I think that may end up having too many variables set. I know there was a user request for being able to build on their own forks so I guess if there's another request for changing the project name then we can look at it?

PromoFaux (approved):
LGTM, plus added DOCKERHUB_NAMESPACE secret to settings

PromoFaux (changes requested):
Can you please apply this same thing to the other two action ymls?

PromoFaux (approved):

PR #940, opened 2021-10-30
Publish images to GHCR alongside DockerHub

ferferga:
## Description
Publishes the same image to GHCR alongside DockerHub.

**Before merging, you must ensure that GITHUB_TOKEN have permissive permissions (the defaults). See [this article](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token= for more information**.

## Motivation and Context
DockerHub is having tighter limits for pulling and GHCR is, honestly, a great alternative where a lot of containers are being published as well. It's also faster to pull from GHCR in my area than DockerHub.

## How Has This Been Tested?
There's no proper way for me to test this as it's only a simple CI change and I don't have token permissions. However, I based my changes in my previous work for GHCR publishing in [Jellyfin Vue](https://github.com/jellyfin/jellyfin-vue/blame/master/.github/workflows/docker.yml) and [Jellyfin's packaging pipelines](https://github.com/jellyfin/jellyfin-metapackages/pull/11)

## Additional Info
I have a script meant to be run as a cron to sync tags that are present in DockerHub but not in GHCR [here](https://gist.github.com/ferferga/7dc7bef353c2f41520a1a42d12f5e41c) (which is really useful when DockerHub limits are reached while pulling the images for syncing). Could be used by some of the maintainers of this project to make GHCR a 1:1 mirror with DockerHub, including older images.

Also, you might want to consider [implementing this another script](https://gist.github.com/ferferga/93ca1ab3056257d05f6e33af0d6ead49) in a cron GitHub Action that prunes images not associated with any tag (something that DockerHub does automatically, but GHCR keeps all the dangling images by default) to keep the package clean. **I can make a PR after this one is merged if this behaviour is desired for you**.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #941, opened 2021-10-31
Remove dangling images from GHCR on schedule

ferferga:
Continuation of #940

## Description
Contrary to DockerHub, GHCR keeps all the images uploaded, even those that are no longer associated to a tag. For example, updating the ``latest`` tags will keep the older image still around in the registry, but untagged.

This new workflow runs every day at 0:00 UTC and removes all the untagged images (aka dangling images)

## Motivation and Context
* There's no point in keeping untagged images at all but GitHub keeps everything by default. CI in general takes a lot of energy, storage and resources. Although those resources are free of charge for us, we don't need them and the environment will be happy with us 😊.

## How Has This Been Tested?
* Ran the script locally.
* Run this same workflow in [my personal repo](https://github.com/ferferga/ferferga)

## Types of changes

## Checklist:

## Needed tokens
A secret ``ACCESS_TOKEN`` (a PAT for the org) must be added in order to remove dangling images, with the following permissions:

* ``delete:packages``
* ``read:packages``

If you prefer another name instead of ``ACCESS_TOKEN`` please let me know.


TonyRL:
FYI, both actions, [GHCR Pruning](https://github.com/marketplace/actions/ghcr-pruning)  and [Container Retention Policy](https://github.com/marketplace/actions/container-retention-policy), support removing dangling images from GHCR.

PromoFaux:
I wonder if https://github.com/pi-hole/.github might be a better place for this (or similar using one of the abovementioned actions)

ferferga:
@PromoFaux The problem is that GH shuts down scheduled actions in a repository after 2 months with no activity, so unless that repo receives activity (not sure what GH considers activity), it won't work properly and will need constant maintenance from your side.

Here's an screenshot of the email I received in one repo where I had no activity:

.github/workflows/maintenance.yml
```diff
@@ -0,0 +1,32 @@
...
+    # Remaining images will be removed in the next scheduled run.
+    continue-on-error: true
+    # Pay attention that the org name is not necessary here, as gh will automatically take the one from the logged in user.
+    # REMEMBER TO KEEP THIS UP TO DATE WITH A LIST OF THE PUBLISHED PACKAGES.
+    strategy:
+      fail-fast: false
+      matrix:
+        container:
+          - 'pihole'
+          - 'ftl-build'
```
PromoFaux, line 22:
```suggestion
          - 'ftl-build'
          - 'docker-pi-hole-base'
```

.github/workflows/maintenance.yml
```diff
@@ -0,0 +1,33 @@
...
+
+    steps:
+      - name: Clone repository 🔽
+        uses: actions/checkout@v2.3.4
+
+      - name: Run deletion script 🗑
+        run: ./delete_ghcr_dangling_images.sh ${{ matrix.container }}
+        working-directory: ./.github/workflows/maintenance
+        env:
+          GITHUB_TOKEN: ${{ secrets.ACCESS_TOKEN }}
```
PromoFaux, line 33:
```suggestion
          GITHUB_TOKEN: ${{ secrets.DELETE_PACKAGE_TOKEN }}
```

PromoFaux (approved):

Issue #922, opened 2021-09-13
Pihole keeps crashing after upgrade


UryDanko:
This is a: Run issue  

## Details

After upgrade to a newest version of Pihole I can't start it anymore. 
Pihole log:
```
pihole           | [s6-init] making user provided files available at /var/run/s6/etc...exited 0.
pihole           | [s6-init] ensuring user provided files have correct perms...exited 0.
pihole           | [fix-attrs.d] applying ownership & permissions fixes...
pihole           | [fix-attrs.d] 01-resolver-resolv: applying...
pihole           | [fix-attrs.d] 01-resolver-resolv: exited 0.
pihole           | [fix-attrs.d] done.
pihole           | [cont-init.d] executing container initialization scripts...
pihole           | [cont-init.d] 20-start.sh: executing...
pihole           |  ::: Starting docker specific checks & setup for docker pihole/pihole
pihole           | Assigning random password: yNlWOoIF
pihole           |
pihole           |   [i] Installing configs from /etc/.pihole...
pihole           |   [i] Existing dnsmasq.conf found... it is not a Pi-hole file, leaving alone!
  [✓] Installed /etc/dnsmasq.d/01-pihole.conf
  [✓] Installed /etc/dnsmasq.d/06-rfc6761.conf
pihole           | sed: cannot rename /etc/pihole/sedjiaqIV: Device or resource busy
pihole           | [cont-init.d] 20-start.sh: exited 4.
pihole           | [cont-finish.d] executing container finish scripts...
pihole           | [cont-finish.d] done.
pihole           | [s6-finish] waiting for services.
pihole           | [s6-finish] sending all processes the TERM signal.
pihole           | [s6-finish] sending all processes the KILL signal and exiting.
pihole           | [s6-init] making user provided files available at /var/run/s6/etc...exited 0.
pihole           | [s6-init] ensuring user provided files have correct perms...exited 0.
pihole           | [fix-attrs.d] applying ownership & permissions fixes...
pihole           | [fix-attrs.d] 01-resolver-resolv: applying...
pihole           | [fix-attrs.d] 01-resolver-resolv: exited 0.
pihole           | [fix-attrs.d] done.
pihole           | [cont-init.d] executing container initialization scripts...
pihole           | [cont-init.d] 20-start.sh: executing...
pihole           |  ::: Starting docker specific checks & setup for docker pihole/pihole
pihole           | Assigning random password: o5lztVC0
pihole           |
[...]

Issue #919, opened 2021-09-12
`REPLY_ADDR*` should be documented


scara:
This is a "**Bug**"

## Details
https://github.com/pi-hole/docker-pi-hole/pull/902 added `REPLY_ADDR*` to improve the FTLDNS configuration when running it within Docker: it should be documented, guessing into the [_Recommended Variables_](https://github.com/pi-hole/docker-pi-hole/tree/1a7bfdc2ee0558d1b184190895a85cdf9fbb7e35#environment-variables) section.

## Related Issues
- [x] I have searched this repository/Pi-hole forums for existing issues and pull requests that look similar

PromoFaux:
Ah yes, it was missed. Would you like to PR it?

scara:
Hi @PromoFaux:
> Would you like to PR it?

Yes, will do.

HTH,
Matteo

dschaper:
FWIW, that variable is set when you use the `ServerIP` environment variable.

dschaper:
https://github.com/pi-hole/docker-pi-hole/pull/902

PromoFaux:
OH. Yes, of course.. that's why it's not in the Readme. Thanks @dschaper 

@scara belay that PR

scara:
Thanks @dschaper!
I'm not familiar with Pi-hole code at all: apologies for the noise 😊.

For the record: `changeFTLsetting` is implemented [here](https://github.com/pi-hole/pi-hole/blob/2673c2c0720fb3871a4c3cae6a3f38f9ccede24e/advanced/Scripts/webpage.sh#L73).

I still wonder if the change in https://github.com/pi-hole/docker-pi-hole/pull/902 should be reported in the README.
Now `ServerIP` is described as:
| Variable | Default | Value | Description |
| -------- | ------- | ----- | ---------- |
| `ServerIP` | unset | `<Host's IP>` | Set to your server's LAN IP, used by web block modes and lighttpd bind address

which should be IMHO updated to tell about FTL REPLY setting will be affected too - and for `ServerIPv6` too.

Is
> lighttpd bind address

still appropriate?

I'm not familiar with the Pi-hole code but the [recent changes](https://github.com/pi-hole/docker-pi-hole/pull/902/files#diff-65a5db1ae8f8e01ebdbbfafcf4e24618ad7cffe036bba92dce1390383e5c73e5L80) suggest to me something like:
``` diff
diff --git a/start.sh b/start.sh
[...]

PR #944, opened 2021-11-21
Customize pihole-FTL.conf with FTLCONF_ prefixed environment variables.

AlexandreRatte:
## Description

Parse all environment variables starting with the `FTLCONF_` prefix and setup the variable's suffix as a `pihole-FTL.conf` customized setting. For example, setting the `FTLCONF_REPLY_ADDR4=10.1.2.3` environment variable would result in a `pihole-FTL.conf` setting of `REPLY_ADDR4=10.1.2.3`.

~~This PR is marked as having a breaking change since the yet undocumented `ServerIP` and `ServerIPv6` variables have been deprecated in favor of provisioning them as `FTLCONF_REPLY_ADDR4` and `FTLCONF_REPLY_ADDR6` respectively.~~

## Motivation and Context

Addresses #922 and #919

There were use cases where `pihole-FTL.conf` could be supplied through a read-only filesystem i.e.: a Kubernetes ConfigMap, and calling the `changeFTLsetting` function on such a read-only file would cause the container to crash loop. This PR is meant to provide the pi-hole administrator all the flexibility that was previously associated with volume-mounting a customized `pihole-FTL.conf`, but now by using environment variables.

## How Has This Been Tested?

- `docker run -it -e FTLCONF_REPLY_ADDR4=10.1.2.3 -e FTLCONF_MAXDBDAYS=7 issue922`
- Shelled into the running container then executed:
```
# cat /etc/pihole/pihole-FTL.conf
PRIVACYLEVEL=0
MAXDBDAYS=7
REPLY_ADDR4=10.1.2.3
#
```

## Types of changes

## Checklist:

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -77,6 +77,8 @@ prepare_configs
 [ -n "${INSTALL_WEB_INTERFACE}" ] && change_setting "INSTALL_WEB_INTERFACE" "$INSTALL_WEB_INTERFACE"
 [ -n "${LIGHTTPD_ENABLED}" ] && change_setting "LIGHTTPD_ENABLED" "$LIGHTTPD_ENABLED"
 [ -n "${DNS_BOGUS_PRIV}" ] && change_setting "DNS_BOGUS_PRIV" "$DNS_BOGUS_PRIV"
+[ -n "${ServerIP}" ] && changeFTLsetting "REPLY_ADDR4" "$ServerIP"
+[ -n "${ServerIPv6}" ] && changeFTLsetting "REPLY_ADDR6" "$ServerIPv6"
 [ -n "${DNS_FQDN_REQUIRED}" ] && change_setting "DNS_FQDN_REQUIRED" "$DNS_FQDN_REQUIRED"
 [ -n "${DNSSEC}" ] && change_setting "DNSSEC" "$DNSSEC"
 [ -n "${REV_SERVER}" ] && change_setting "REV_SERVER" "$REV_SERVER"
--- a/start.sh
+++ b/start.sh
@@ -76,8 +76,6 @@ prepare_configs
 [ -n "${INSTALL_WEB_SERVER}" ] && change_setting "INSTALL_WEB_SERVER" "$INSTALL_WEB_SERVER"
 [ -n "${INSTALL_WEB_INTERFACE}" ] && change_setting "INSTALL_WEB_INTERFACE" "$INSTALL_WEB_INTERFACE"
 [ -n "${LIGHTTPD_ENABLED}" ] && change_setting "LIGHTTPD_ENABLED" "$LIGHTTPD_ENABLED"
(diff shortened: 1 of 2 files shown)


PromoFaux:
I've not tested this functionally, yet, but I like the idea a lot. I am wondering (out loud) if we can do something similar for `setupVars.conf` keypairs, so `SETUPVARS_` env variables, and what this would look like. 

I also need to check a few things, but we might be able to keep the `ServerIP` ones around so that this does not become a breaking change (although I'm all for breaking changes if they are good and sensible like this)

AlexandreRatte:
Yes, implementing something similar for `setupVars.conf` would probably clean things up rather nicely. However, in light of #922 causing issues to folks who are supplying `pihole-FTL.conf` through a read-only volume mount due to the lack of customization capabilities by other means, I would suggest attacking `setupVars.conf` outside of this PR.

Let me know of your thoughts, I'm happy to discuss and make amendments per your review.

start.sh
```diff
@@ -76,8 +76,6 @@ prepare_configs
 [ -n "${INSTALL_WEB_SERVER}" ] && change_setting "INSTALL_WEB_SERVER" "$INSTALL_WEB_SERVER"
 [ -n "${INSTALL_WEB_INTERFACE}" ] && change_setting "INSTALL_WEB_INTERFACE" "$INSTALL_WEB_INTERFACE"
 [ -n "${LIGHTTPD_ENABLED}" ] && change_setting "LIGHTTPD_ENABLED" "$LIGHTTPD_ENABLED"
-[ -n "${ServerIP}" ] && changeFTLsetting "REPLY_ADDR4" "$ServerIP"
```
PromoFaux, line 79: For ease of updating, I think we should leave these two lines in - but add a note to the README/release notes that they will be removed in a later version, and should be replaced with `FTLCONF_REPLY_ADDR4/6` 

I'll look at untangling `ServerIP` from the rest of the codebase at some point:
image

PromoFaux (changes requested):
For ease of updating, I think we should leave these two lines in - but add a note to the README/release notes that they will be removed in a later version, and should be replaced with `FTLCONF_REPLY_ADDR4/6` 

I'll look at untangling `ServerIP` from the rest of the codebase at some point:
image

start.sh
PromoFaux, line 79: For ease of updating, I think we should leave these two lines in - but add a note to the README/release notes that they will be removed in a later version, and should be replaced with `FTLCONF_REPLY_ADDR4/6` 

I'll look at untangling `ServerIP` from the rest of the codebase at some point:
image

PromoFaux (changes requested):

start.sh
AlexandreRatte: Ok, I've reintroduced parsing of the `ServerIP*` variables and made changes to `README.md` which should highlight their deprecated nature, while also explicitly mentioning that `FTLCONF_REPLY_ADDR4` is recommended.

PromoFaux (approved):

Issue #947, opened 2021-11-29
docker-compose v2.1.1 exposes problem with trailing slashes in PiHole compose file examples


Paraphraser:
This is a: Bug  

## Details

To be clear, I am not reporting a bug in PiHole. I am reporting a problem with PiHole **documentation** which has been exposed by a change in how up-to-date versions of docker-compose process `volumes` directives in compose files. 

The example compose files for PiHole provided here on GitHub and on DockerHub include:

```
volumes:
- ./volumes/pihole/etc-pihole/:/etc/pihole/
- ./volumes/pihole/etc-dnsmasq.d/:/etc/dnsmasq.d/
```

Under version 2.1.1 of docker-compose, the trailing slashes on those path components cause an error when the left-hand-sides of those `volumes` directives do not exist when `docker-compose up -d` is invoked. The error message is:

```
Error response from daemon: invalid mount config for type "bind": bind source path does not exist:
```

Removing the trailing slashes cures the problem.

This problem was **not** apparent with 1.x versions of docker-compose. It *may* have been apparent with 2.x versions of docker-compose prior to 2.1.1 but I have not tested to determine precisely when this problem first surfaced. All I can really say for certain is that it is there **now**, is **consistent** and is **repeatable**.

Because the problem is only triggered by the combination of up-to-date docker-compose **plus** missing left-hand-sides of `volumes` directives, the people most likely to be affected are new users on first install and those trying to debug PiHole problems by erasing persistent storage. It will not affect a PiHole user who simply upgrades their version of docker-compose.

PiHole is not the only container that is going to bang into this. I found 5 problematic containers in [SensorsIot/IOTstack](https://github.com/SensorsIot/IOTstack), including PiHole. I have submitted Pull Requests on SensorsIot/IOTstack. I'm opening this issue because you might want to inform your wider audience by changing the PiHole documentation on GitHub and DockerHub to reflect this new externally-imposed reality.

## Related Issues

- [✔︎] I have searched this repository/Pi-hole forums for existing issues and pull requests that look similar 

## How to reproduce the issue 

1. Environment data

  * Operating System: Raspbian (Buster or Bullseye)
  * Hardware: Raspberry Pi 3B+, 4B (4GB) or PiZeroW2 (but probably all RPi)
  * Kernel Architecture: Both armv7l and aarch64
  * Docker Install Info and version: 
    - Software source: Official Docker "convenience script" - `curl -fsSL https://get.docker.com | sudo sh`
[...]

PR #949, opened 2021-12-09
Trailing slashes on volumes fail on newer versions of docker-compose

Paraphraser:
## Description, Motivation and Context

See [Issue 947](https://github.com/pi-hole/docker-pi-hole/issues/947) for the background.

This PR proposes removing trailing slashes from all documentation, examples and scripts.

## How Has This Been Tested?

Tested with docker-compose 2.1.1:

* **with** trailing slashes present on volumes statements = fails
* **without** trailing slashes present on volumes statements = succeeds

## Types of changes

## Checklist:

## Note:

The documentation on DockerHub also has trailing slashes. I don't know if the GitHub README.md automatically propagates to DockerHub. If it does not, DockerHub should also be updated.


PromoFaux:
There is nothing in place at the moment to auto update dockerhub docs when the README.md is updated here. I looked at incorporating an action to do this earlier, but there is a hard limit via the docker hub API of 25000 bytes (README currently comes in at around 28000 bytes 😳)

One thing I would like to do is pare down the Readme on this repo, and move it all to a dedicated section of https://docs.pi-hole.net... I just haven't found the time/energy to yet!

PromoFaux:
Also, sorry to be a pain - I know it's "only" documentation, but could you please add a DCO sign off to your commit per the instructions here? https://github.com/pi-hole/docker-pi-hole/pull/949/checks?check_run_id=4490219562

That will please the bot

Paraphraser:
I have absolutely no idea what that all means. Not the language "DCO", and not the requisite commands. All Greek, so to speak.

I still had the clone+branch for the PR so I followed the instructions but it barfed with

37ADC345-9453-4FC8-9702-4B6FCFADF8B5

On a hunch I did a pull origin branch name and tried again. It seemed to work. I hope that was sufficient because I'd be at a loss to know what to do next.

PromoFaux:
It's OK, sometimes it can be a bit fiddly when force pushing gets involved - but I can see at least one of the commits has your signoff, so I can manaully override the bot in this case. For more info on DCO: https://probot.github.io/apps/dco/

PromoFaux (approved):

Paraphraser:
In case I ever want to submit another PR for PiHole, can I please ask a follow-up question?

Normally, I end any PR process with a `git commit -S` which results in my YubiKey wanting a tap to authorise the GPG signing, and then shows up on GitHub as "This commit was signed with the committer’s verified signature". Until now, I thought that was sufficient.

If I'm reading the git doco correctly in the context of the repair commands needed for DCO, it looks to me that I *might* have avoided the entire problem with this PR if I had simply added `-s` to the original command, as in:

```
git commit -s -S
``

Is that all there is to it or am I missing something?

> Old dude here. Grew up with CVS and subversion (not to mention something called UPDATE on Control Data mainframes back-in-the-day). Git and voodoo are right next to each other in my tech lexicon.

PromoFaux:
Ha, don't worry about it - I'm still finding out new things about git every day, myself. I cheat in a lot of cases and use gui tools

`gpg` signing is separate. (And a hard requirement for core developers on the project) For example, I have an option in my `.gitconfig` file that ensures all commits are gpg signed, and indeed I get prompted to enter my yubi's pin. I believe that is also the same as `-S`

For the DCO bot, we're just looking for the `Signed-off by:` part at the end of a commit message, which as you note is generated by the `-s` flag. It's easier to get new contributors (who may never even have touched source control before) to sign off that way, than it is to get them to set up gpg 🙃

Issue #952, opened 2021-12-23
duplicate PIHOLE_DOCKER_TAG after upgrade


mfschumann:
### Versions

- Pi-hole:  v5.7
- AdminLTE:  v5.9
- FTL:  v5.12

### Platform

- OS and version:  Fedora 35

- Platform:  Docker

### Expected behavior

Pihole should work after upgrading the docker container to the newest available version of the pihole image.

### Actual behavior / bug

The upgrade adds an additional 'PIHOLE_DOCKER_TAG' line to the config in lighttpd/conf-enabled/15-fastcgi-php.conf. That causes pihole to fail starting because of the duplicate PIHOLE_DOCKER_TAG.

### Steps to reproduce

1. Set up pihole docker.
2. Delete the docker container, but keep the config.
3. Set up a newer version of the pihole docker container, using the same config files as before.

## Debug Token

- URL:  

### Screenshots

_If applicable, add screenshots to help explain your problem._

### Additional context

My current 15-fastcgi-php.conf:

```
# -*- depends: fastcgi -*-
# /usr/share/doc/lighttpd/fastcgi.txt.gz
# http://redmine.lighttpd.net/projects/lighttpd/wiki/Docs:ConfigurationOptions#mod_fastcgi-fastcgi

## Start an FastCGI server for php (needs the php5-cgi package)
fastcgi.server += ( ".php" => 
	((
		"bin-path" => "/usr/bin/php-cgi",
		"socket" => "/var/run/lighttpd/php.socket",
		"max-procs" => 1,
		"bin-environment" => ( 
			"PIHOLE_DOCKER_TAG" => "2021.12",
			"CORS_HOSTS" => "",
			"VIRTUAL_HOST" => "dns.mfs.name",
			"PHP_ERROR_LOG" => "/var/log/lighttpd/error.log",
			"ServerIP" => "192.168.0.50",
			"PHP_FCGI_CHILDREN" => "4",
			"PHP_FCGI_MAX_REQUESTS" => "10000"
		),
		"bin-copy-environment" => (
			"PATH", "SHELL", "USER"
		),
		"broken-scriptfilename" => "enable"
	))
)
```
Immediately after the upgrade from 2021.11 to 2021.12, an additional line reading
```
"PIHOLE_DOCKER_TAG" => "2021.11",
```
was present directly below the line containing PIHOLE_DOCKER_TAG in the current config. I removed the additional line to make pihole start again.

jpspgd:
[...]

PR #957, opened 2021-12-28
Fix duplicate php env config values after upgrades and config changes

mfschumann:
## Description
I changed the behavior of setup_php_env such that it changes lines that are already present instead of adding them always to PHP_ENV_CONFIG.

## Motivation and Context
The previous behavior of setup_php_env lead to duplicate config entries upon upgrading the container image (i.e. changes to PIHOLE_DOCKER_TAG) or changes to any of the other config parameters. This has been reported in #952.

## How Has This Been Tested?
I ran setup_php_env with different php env config files and verified that the config files were OK.

## Types of changes

## Checklist:

### DIFF
--- a/bash_functions.sh
+++ b/bash_functions.sh
@@ -198,23 +198,17 @@ setup_php_env() {
     if [ -z "$VIRTUAL_HOST" ] ; then
       VIRTUAL_HOST="$ServerIP"
     fi;
-    local vhost_line="\t\t\t\"VIRTUAL_HOST\" => \"${VIRTUAL_HOST}\","
-    local corshosts_line="\t\t\t\"CORS_HOSTS\" => \"${CORS_HOSTS}\","
-    local serverip_line="\t\t\t\"ServerIP\" => \"${ServerIP}\","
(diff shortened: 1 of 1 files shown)


PromoFaux:
Thanks for this, logic looks about right, I will aim to test it by the end of the week (or year, even!)

PromoFaux (approved):
I like having this in a for loop rather than a line for each - it is a little easier to read.

Having tested it, I can see it is still produces the correct results on a fresh container (i.e - one where the file `/etc/lighttpd/conf-enabled/15-fastcgi-php.conf` has not been persisted on the host system)

kwilczynski:
Hi @mfschumann,

Thank for you for fixing this! It was a source of some annoyance during upgrades for me! Kudos!

PromoFaux:
@kwilczynski  see also: https://github.com/pi-hole/docker-pi-hole/issues/952#issuecomment-1002292608

kwilczynski:
Hi @PromoFaux,

Had a look. Does not apply to me. Thank you for letting me know, though!

PromoFaux:
Interesting, you should only have been coming up against this if you were persisting that file. We'll find out on the next release (imminent!)

PR #973, opened 2022-01-07
Prevent logspam when restarting FTL (Error from kill command)

PromoFaux:
## Description

Waits until the process is dead before attempting to start it again. Prevents this:

image

## Motivation and Context

The logspam is annoying

## How Has This Been Tested?

I restarted FTL within the container several times. It no longer prints the warning about the usage of `kill`

image

## Types of changes

## Checklist:

### DIFF
--- a/s6/service
+++ b/s6/service
@@ -2,29 +2,49 @@
 # This script patches all service commands into the appropriate s6- commands
 # pi-hole upstream scripts need a 'service' interface. why not systemd? docker said so.
 start() {
-    s6-svc -wU -u -T2500 /var/run/s6/services/$service
+  s6-svc -wU -u -T2500 /var/run/s6/services/$service
 }
 
 stop() {
-    s6-svc -wD -d -T2500 /var/run/s6/services/$service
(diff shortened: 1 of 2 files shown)


PromoFaux:
Unintended, but pleasing effect of this is that this also stops pihole-FTL starting several times on container start:

Left is `:dev`, right is current `:latest`
image

s6/service
```diff
@@ -2,29 +2,49 @@
...
+
+  # Only attempt to stop the service if it is already running
+  if [ -n "$pid" ]; then
     stop
+
+    # Loop until we are certain that the process has been stopped
+    while test -d /proc/$pid; do
+      sleep 0.2
+    done
+  fi
```
dschaper, line 26: Might want something like an upper limit on time so that you don't accidentally infinite loop yourself.

PromoFaux: I'm trying to think of what could go wrong to get us stuck in here - i.e FTL not stopping for whatever reason - buy you're probably right, more guards are better!

PR #977, opened 2022-01-08
Revert "Remove dangling images from GHCR on schedule"

PromoFaux:
Reverts pi-hole/docker-pi-hole#941

Doesn't appear to work work for Org accounts (should have caught that when reading the script)- will use one of the recommended actions


PromoFaux:
FP'd `dev` to 0baf7b63d926743f4a53afed43701c68c1de802b to tidy up merge then immediate reversion.

PromoFaux:
@ferferga - The API endpoints are different for org accounts, which is why it failed. I also have looked at https://github.com/vlaurin/action-ghcr-prune, but that's also failing, and I don't have much time to dig into it just now.

https://github.com/pi-hole/.github/runs/4748932595?check_suite_focus=true

Surprised GH don't automatically prune dangling images themselves, tbh. No matter - that's up to them!

PR #990, opened 2022-01-26
[Experimental] Add ability to Redirect lighttpd logs to stdout

edgd1er:
Feature: upon env value, redirect lighttpd logs to stdout

## Description

if the WEBLOGS_STDOUT is not 0 (0 is default value), then two fifo pipes are created at the same place the original files should be.
existing files are removed.
Two services are started to echo logs to stdout.
Two warnings are issued. I didn't find a way to discard them.
```
docker-pi-hole-pihole-1  | s6-svc: warning: /var/run/s6/services/lighttpd-access-log/notification-fd not present - ignoring request for readiness notification
docker-pi-hole-pihole-1  | s6-svc: warning: /var/run/s6/services/lighttpd-error-log/notification-fd not present - ignoring request for readiness notification
docker-pi-hole-pihole-1  | Redirecting FIFO access log to stdout
docker-pi-hole-pihole-1  | Redirecting FIFO error log to stdout
docker-pi-hole-pihole-1  | 2022-01-26 10:49:59: server.c.1513) server started (lighttpd/1.4.59)
docker-pi-hole-pihole-1  | 1643190629|localhost:8053|GET /admin/ HTTP/1.1|200|17014
docker-pi-hole-pihole-1  | 1643190629|localhost:8053|GET /admin/style/vendor/SourceSansPro/SourceSansPro.css?v=1643190585 HTTP/1.1|200|3516
docker-pi-hole-pihole-1  | 1643190629|localhost:8053|GET /admin/style/vendor/bootstrap/css/bootstrap.min.css?v=1643190585 HTTP/1.1|200|121457
docker-pi-hole-pihole-1  | 1643190629|localhost:8053|GET /admin/style/vendor/datatables.min.css?v=1643190585 HTTP/1.1|200|4672
```

## Motivation and Context

Sometimes, it's convenient to have lighttpds access and error logs in container's logs.
A simple docker compose logs will show web server logs.

## How Has This Been Tested?

When I start the container, I can see logs from the admin website, having added `WEBLOGS_STDOUT=1` to the env vars.

Changes are limited to container start up scripts.

## Types of changes

## Checklist:

### DIFF
--- a/s6/debian-root/etc/services.d/lighttpd-access-log/finish
+++ b/s6/debian-root/etc/services.d/lighttpd-access-log/finish
@@ -1,0 +1,9 @@
+#!/usr/bin/with-contenv bash
+
+s6-echo "Stopping lighttpd-access-log"
+pid=$(ps -C cat -o pid=,args= |grep -oP "([0-9]+).+access\.log" |cut -f1 -d" ")
+if [[ -n ${pid} ]]; then
+  echo "Killing ${pid}"
+  kill -9 ${pid}
+fi
+s6-echo "Stopped lighttpd-access-log"
--- a/s6/debian-root/etc/services.d/lighttpd-access-log/finish
+++ b/s6/debian-root/etc/services.d/lighttpd-access-log/finish
@@ -3,7 +3,6 @@
 s6-echo "Stopping lighttpd-access-log"
 pid=$(ps -C cat -o pid=,args= |grep -oP "([0-9]+).+access\.log" |cut -f1 -d" ")
 if [[ -n ${pid} ]]; then
-  echo "Killing ${pid}"
   kill -9 ${pid}
 fi
 s6-echo "Stopped lighttpd-access-log"
--- a/s6/debian-root/etc/services.d/lighttpd-access-log/run
+++ b/s6/debian-root/etc/services.d/lighttpd-access-log/run
@@ -1,0 +1,5 @@
+#!/usr/bin/with-contenv bash
+
+s6-echo "Redirecting FIFO access log to stdout"
+
+s6-setuidgid www-data cat /var/log/lighttpd/access.log 2>&1
--- a/s6/debian-root/etc/services.d/lighttpd-access-log/run
+++ b/s6/debian-root/etc/services.d/lighttpd-access-log/run
@@ -1,5 +1,5 @@
 #!/usr/bin/with-contenv bash
 
-s6-echo "Redirecting FIFO access log to stdout"
+s6-echo "Starting lighttpd-access-log"
 
 s6-setuidgid www-data cat /var/log/lighttpd/access.log 2>&1
--- a/s6/debian-root/etc/services.d/lighttpd-error-log/finish
+++ b/s6/debian-root/etc/services.d/lighttpd-error-log/finish
@@ -3,7 +3,6 @@
 s6-echo "Stopping lighttpd-error-log"
 pid=$(ps -C cat -o pid=,args= |grep -oP "([0-9]+).+error\.log" |cut -f1 -d" ")
 if [[ -n ${pid} ]]; then
-  echo "Killing ${pid}"
   kill -9 ${pid}
 fi
 s6-echo "Stopped lighttpd-error-log"
--- a/s6/debian-root/etc/services.d/lighttpd-error-log/finish
+++ b/s6/debian-root/etc/services.d/lighttpd-error-log/finish
@@ -1,0 +1,9 @@
+#!/usr/bin/with-contenv bash
+
+s6-echo "Stopping lighttpd-error-log"
+pid=$(ps -C cat -o pid=,args= |grep -oP "([0-9]+).+error\.log" |cut -f1 -d" ")
+if [[ -n ${pid} ]]; then
+  echo "Killing ${pid}"
+  kill -9 ${pid}
+fi
+s6-echo "Stopped lighttpd-error-log"
--- a/s6/debian-root/etc/services.d/lighttpd-error-log/run
+++ b/s6/debian-root/etc/services.d/lighttpd-error-log/run
@@ -1,5 +1,5 @@
 #!/usr/bin/with-contenv bash
 
-s6-echo "Redirecting FIFO error log to stdout"
+s6-echo "Starting lighttpd-error-log"
 
 s6-setuidgid www-data cat /var/log/lighttpd/error.log 2>&1
--- a/s6/debian-root/etc/services.d/lighttpd-error-log/run
+++ b/s6/debian-root/etc/services.d/lighttpd-error-log/run
@@ -1,0 +1,5 @@
+#!/usr/bin/with-contenv bash
+
+s6-echo "Redirecting FIFO error log to stdout"
+
+s6-setuidgid www-data cat /var/log/lighttpd/error.log 2>&1
--- a/s6/debian-root/etc/services.d/lighttpd/finish
+++ b/s6/debian-root/etc/services.d/lighttpd/finish
@@ -1,6 +1,6 @@
 #!/usr/bin/with-contenv bash
 
 s6-echo "Stopping lighttpd"
-killall -9 lighttpd
 service lighttpd-access-log stop
 service lighttpd-error-log stop
+killall -9 lighttpd
--- a/s6/debian-root/etc/services.d/lighttpd/finish
+++ b/s6/debian-root/etc/services.d/lighttpd/finish
@@ -2,3 +2,5 @@
 
 s6-echo "Stopping lighttpd"
 killall -9 lighttpd
+service lighttpd-access-log stop
+service lighttpd-error-log stop
--- a/s6/debian-root/etc/services.d/lighttpd/run
+++ b/s6/debian-root/etc/services.d/lighttpd/run
@@ -2,11 +2,23 @@
 
 s6-echo "Starting lighttpd"
 
-# Touch log files to ensure they exist (create if non-existing, preserve if existing)
-touch /var/log/lighttpd/access.log /var/log/lighttpd/error.log
+if [[ 1 -eq ${WEBLOGS_STDOUT:-0} ]]; then
+  #lighthttpd cannot use /dev/stdout https://redmine.lighttpd.net/issues/2731
+  rm -f /var/log/lighttpd/access.log
+  mkfifo -m 600 /var/log/lighttpd/access.log
+  rm -f /var/log/lighttpd/error.log
+  mkfifo -m 600 /var/log/lighttpd/error.log
+  chown -R www-data:www-data /var/log/lighttpd
+  service lighttpd-access-log start
+  service lighttpd-error-log start
+  sleep 2
+else
+  # Touch log files to ensure they exist (create if non-existing, preserve if existing)
+  touch /var/log/lighttpd/access.log /var/log/lighttpd/error.log
 
-# Ensure that permissions are set so that lighttpd can write to the logs
-chown -R www-data:www-data /var/log/lighttpd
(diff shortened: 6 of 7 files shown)


PromoFaux:
Nice, thanks for this contribution - I will aim to look at it later after work (if I can!)

Please sign off your commits to appease the DCO bot :)

edgd1er:
Damned, I always forget that DCO stuff. done.

PromoFaux:
Thanks, it seems to work well... a couple of things I've been playing with:
 
  - remove the additional services, and instead of having the `lighttpd` service start the extra two services, replace those calls with a direct, e.g `s6-setuidgid www-data cat /var/log/lighttpd/access.log 2>&1 &` (note the final `&` so that it is non blocking). This seems to work, and prevents that additional output warning.
  - Doing this, but also for `pihole-FTL.log` and `pihole.log` (each as separate variables because `pihole.log` is BUSY)

For that first point, the whole section in the `run` part of `lighttpd`'s "service" becomes:

```bash
if [[ 1 -eq ${WEBLOGS_STDOUT:-0} ]]; then
  #lighthttpd cannot use /dev/stdout https://redmine.lighttpd.net/issues/2731
  rm -f /var/log/lighttpd/access.log
  mkfifo -m 600 /var/log/lighttpd/access.log
  rm -f /var/log/lighttpd/error.log
  mkfifo -m 600 /var/log/lighttpd/error.log
  chown -R www-data:www-data /var/log/lighttpd

  s6-setuidgid www-data cat /var/log/lighttpd/access.log 2>&1 &
  s6-setuidgid www-data cat /var/log/lighttpd/error.log 2>&1 &
fi
```

PromoFaux:
Something I'm trying to hack in/see if it's possible...  it would be nice to maybe prefix each line of the access/error log with the name of the log file, perhaps. 

I don't know if a) this is possible, or b) it's even desirable. Just thinking about tidying up output is all :)

PromoFaux:
The pihole-FTL.log is a total aside, by the way, and can come in another PR all together - I'm just playing at the moment!

edgd1er:
Concerning the log service, the two log services, the idea was to be able to restart them if needed, not sure it would be very helpful though. the second point was to have rather direct parenting link between pid 0 and cat fifo through s6-services. I don't know how well backgrounded commands are still linked to the parent. If cat are killed in a improper way, fifo files may keep data that would be fetched at next container's start ( just a guess, i'm not familiar with fifo neither)

As for prefixing the log with its "origin", this should occur before sending to the fifo, so through log formating in the application (pihole FTL, lighttpd). s6-logservice might help, but I must admit, I'm having difficulties to understand S6 documentation.

PromoFaux:
Something interesting I came across when trying to reproduce with `pihole-FTL.log`... in that case it needs to be `tail -f` instead of `cat` - I don't really know why... magic low level stuff I don't understand would be my guess...

PromoFaux:
Doing some more playing with this, is there a reason they are being created as named pipes rather than just `tail -f`ing the existing files?

edgd1er:
Using pipes, logs are "consumed". if you write them (no fifo) and cat them, you will have to information, in the log file and in the container's log. At beginning, I was willing to have either the log file or the container's log. a Third way, would be to have both of them, but you'll have to handle logs rotation and container's log through docker log's management.
WEBLOGS could have three values: STDOUT, LOGS, BOTH.

PromoFaux:
> Two warnings are issued. I didn't find a way to discard them.

[Might be something useful here](https://skarnet.org/software/s6/s6-svc.html#:~:text=%2DwU%C2%A0%3A%20s6%2Dsvc%20will%20not%20exit%20until%20the%20service%20is%20up%20and%20ready%20as%20notified%20by%20the%20daemon%20itself.%20If%20the%20service%20directory%20does%20not%20contain%20a%20notification%2Dfd%20file%20to%20tell%20s6%2Dsupervise%20to%20accept%20readiness%20notification%2C%20s6%2Dsvc%20will%20print%20a%20warning%20and%20act%20as%20if%20the%20%2Dwu%20option%20had%20been%20given%20instead.)

PromoFaux:
I think, really, the `-wU` is unnessacery here. Again, we can deal with that in another patch/PR - just noting it down somewhere to remember!

s6/debian-root/etc/services.d/lighttpd-access-log/run
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/with-contenv bash
+
+s6-echo "Redirecting FIFO access log to stdout"
+
+s6-setuidgid www-data cat /var/log/lighttpd/access.log 2>&1
```
PromoFaux, line 5:
```suggestion
s6-setuidgid www-data cat /var/log/lighttpd/access.log 2>&1

```

New line please

s6/debian-root/etc/services.d/lighttpd/run
```diff
@@ -2,11 +2,23 @@
...
-# Touch log files to ensure they exist (create if non-existing, preserve if existing)
-touch /var/log/lighttpd/access.log /var/log/lighttpd/error.log
+if [[ 1 -eq ${WEBLOGS_STDOUT:-0} ]]; then
+  #lighthttpd cannot use /dev/stdout https://redmine.lighttpd.net/issues/2731
+  rm -f /var/log/lighttpd/access.log
+  mkfifo -m 600 /var/log/lighttpd/access.log
+  rm -f /var/log/lighttpd/error.log
+  mkfifo -m 600 /var/log/lighttpd/error.log
+  chown -R www-data:www-data /var/log/lighttpd
+  service lighttpd-access-log start
```
PromoFaux, line 12: I'm still unsure of the best way to handle these.

Just for fun, exec into to the container and manually stop/start these services. Stopping does nothing and the two `cat` processes stay alive. Starting them again adds n number of additional `cat` processes.

Also restarting just the `lighttpd` service causes a loop:

image

PromoFaux (changes requested):

s6/debian-root/etc/services.d/lighttpd/run
edgd1er: adding these lines into lighttpd's finish fixes the problem.
```
service lighttpd-access-log stop
service lighttpd-error-log stop
killall -9 cat
```

s6/debian-root/etc/services.d/lighttpd/finish
```diff
@@ -2,3 +2,5 @@
 
 s6-echo "Stopping lighttpd"
 killall -9 lighttpd
+service lighttpd-access-log stop
+service lighttpd-error-log stop
```
PromoFaux, line 6: This is closer, at least, but something odd in the output when stopping `lighttpd`:

```
pihole    | Stopping lighttpd
pihole    | Stopping lighttpd-error-log
pihole    | Stopping lighttpd-access-log
pihole    | Stopped lighttpd-error-log
pihole    | Stopped lighttpd-access-log <---------expected up to here
pihole    | Redirecting FIFO error log to stdout <------ wait, error log service started again?
pihole    | Stopping lighttpd-error-log <----- only to get stopped again
pihole    | Killing 783 <--- expected per your addition in the finish scripts for each service
pihole    | Stopped lighttpd-error-log <---- expected
```

PromoFaux: Ah, worked it out.. the calls to stop the two services need to happen _before_ `lighttpd` is killed so:

```
service lighttpd-access-log stop
service lighttpd-error-log stop
killall -9 lighttpd
```

s6/debian-root/etc/services.d/lighttpd-access-log/run
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/with-contenv bash
+
+s6-echo "Redirecting FIFO access log to stdout"
```
PromoFaux, line 3:
```suggestion
s6-echo "Starting lighttpd-access-log"
```

s6/debian-root/etc/services.d/lighttpd-access-log/finish
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/with-contenv bash
+
+s6-echo "Stopping lighttpd-access-log"
+pid=$(ps -C cat -o pid=,args= |grep -oP "([0-9]+).+access\.log" |cut -f1 -d" ")
+if [[ -n ${pid} ]]; then
+  echo "Killing ${pid}"
```
PromoFaux, line 6:
```suggestion
```
We probably don't need to be this verbose here

s6/debian-root/etc/services.d/lighttpd-error-log/finish
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/with-contenv bash
+
+s6-echo "Stopping lighttpd-error-log"
+pid=$(ps -C cat -o pid=,args= |grep -oP "([0-9]+).+error\.log" |cut -f1 -d" ")
+if [[ -n ${pid} ]]; then
+  echo "Killing ${pid}"
```
PromoFaux, line 6:
```suggestion
```

As above, less verbosity

s6/debian-root/etc/services.d/lighttpd-error-log/run
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/with-contenv bash
+
+s6-echo "Redirecting FIFO error log to stdout"
```
PromoFaux, line 3:
```suggestion
s6-echo "Starting lighttpd-error-log"
```

s6/debian-root/etc/services.d/lighttpd/run
```diff
@@ -2,11 +2,23 @@
 
 s6-echo "Starting lighttpd"
 
-# Touch log files to ensure they exist (create if non-existing, preserve if existing)
-touch /var/log/lighttpd/access.log /var/log/lighttpd/error.log
+if [[ 1 -eq ${WEBLOGS_STDOUT:-0} ]]; then
+  #lighthttpd cannot use /dev/stdout https://redmine.lighttpd.net/issues/2731
+  rm -f /var/log/lighttpd/access.log
```
PromoFaux, line 7: I wonder if it is also worth putting a check in here to only rm the file if it is not a pipe, i.e `if [[ ! -p /var/log/lighttpd/access.log ]] remove it` (and same for the error log)

Might be overkill!

s6/debian-root/etc/services.d/lighttpd/finish
PromoFaux: Also if we're going to have a `Stopped ` message for the other two services, perhaps we should add an `s6-echo "Stopped lighttpd" here?

PromoFaux (changes requested):

s6/debian-root/etc/services.d/lighttpd/finish
edgd1er: ok you have found the solution. I was wondering also with that order .... 
I also tried to replace bash command with s6 command
rm -rf => s6-rmrf
mkfifo => s6-mkfifo
chown =>s6-chown
the latter is tricky because it requires uid gid instead of a name.
not sure what is better s6 cmd or bash cmd.

PromoFaux: > not sure what is better s6 cmd or bash cmd.

At this stage, "whatever works" is good enough for me. I started looking at `3.0.0.0` last night, its a learning curve!

edgd1er: I had a look also ;) . all services should be re desingned as the new s6 way ( s6-rc.d). At the moment, stumbled on 2 errors (cont-profile.d missing but not to be create or else $etc is wrongly set to /etc/s6-profile.d/env or something alike. The container starts but no services are running .....

PromoFaux (approved):

PR #991, opened 2022-01-26
Touch/chown lighttpd access/error logs on service start

PromoFaux:
## Description

Playing about with some alternatives to https://github.com/pi-hole/docker-pi-hole/pull/990, I noticed that if we didn't create FIFO pipes, but instead just `tail -f` existing log files, they need to actually exist first...

## Types of changes

## Checklist:


edgd1er (commented):
chown -R www-data:www-data /var/log/lighttpd would also set perms on dir, which sometimes is needed.

PromoFaux:
Thanks, that makes sense actually. I've tweaked it

PR #998, opened 2022-02-11
S6 v3: move to last version of s6-overlay

edgd1er:
## Description

Prepare pi-hole for the next major upgrade: s6-overlay v3.
You may wish to create a new branch for that upgrade.
A new version version is planned next week or so. I would fix the S6_GLOBAL_PATH var being ignored + many other fixes.

All legacy services/cont-init.d/finish.d are transformed into new features:
- oneshot for init
- longshot for services

Dependencies were set when possible.

## Motivation and Context

Keep up with the maintained version

No issue are planned to be fixed nor revealed with this change.

## How Has This Been Tested?

I tested it on my running system

- checked lighttpd logs
- checked dns resolution
- pihole -g
- pihole status

changes affect /etc/s6-overlay/ and s6-overlay install dirs (/command, ....)

not test on other system than a x86_64. will do !
I just noticed that I also pushed the base image change. (debian-bullseye). I guess I should revert that.

WIP: i need to adapt the dockerfile to be less different from yours

## Types of changes

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,7 @@ var-log/
 
 # WIP/test stuff
 doco.yml
+/.idea/
+
+*.iml
+/logs/
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,47 +1,57 @@
+# syntax=docker/dockerfile:1.3
 ARG PIHOLE_BASE
-FROM "${PIHOLE_BASE:-ghcr.io/pi-hole/docker-pi-hole-base:bullseye-slim}"
+# =ghcr.io/pi-hole/docker-pi-hole-base:bullseye-slim
+FROM ${PIHOLE_BASE}
 
 ARG PIHOLE_DOCKER_TAG
 ENV PIHOLE_DOCKER_TAG "${PIHOLE_DOCKER_TAG}"
+ARG TARGETPLATFORM
+ARG BUILDPLATFORM
 
-ENV S6_OVERLAY_VERSION v2.1.0.2
+ARG aptCacher
+ARG S6_OVERLAY_VERSION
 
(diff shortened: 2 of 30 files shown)


PromoFaux:
Ah, wow. thanks for looking into this!

edgd1er:
add apt cache with the help of buildkit (required)
fix redirect logs to stdout on/off
still an issue on armhf: pihole-FTL permission denied.

PromoFaux:
Just FYI - If you want to run the tests locally, btw, you can run `./gh-actions-test.sh`

edgd1er:
> Just FYI - If you want to run the tests locally, btw, you can run `./gh-actions-test.sh`

Unfortunately, I'm using buildkit to speed up building process. the actual test process does not setup buildkit. 
I think you can close the PR as I'm going in a direction too much different from the projet ( multistage build, buildkit usage, build cache mount ). At the moment, overlay V3 is running on two amd64 and two armhf without problems. I haven't conducted performance tests though.
Changing the test process, the build process and the overlay version in one PR is a lot. I guess you 'd like to split the subject in different PR.
If you are open to the idea of changing test and build process, i can have a look to it.

PromoFaux:
The actual final build & publish step uses `buildx` (via github actions):

https://github.com/pi-hole/docker-pi-hole/blob/0018e46a288cf7a6e9544d1d2fd11d3ff4d37f40/.github/workflows/test-and-build.yaml#L26-L85

The `./gh-actions-test` can likely stand to be looked at in more detail at some point. Initially we were doing all sorts of stuff with python scripts to create multiarch builds.. in the end it was easier to go with native docker build stuff with GHA.

PR #1014, opened 2022-03-09
Support docker service names and links for PIHOLE_DNS_

haxwithaxe:
Added the ability to use docker service names or links in `PIHOLE_DNS_`.

## Description
In `start.sh` I added a check for valid domains to the block in the loop over `PIHOLE_DNS_`. The valid domains are resolved and the IP addresses are added to the configs in the same way as DNS servers specified as IPs. I also slightly rearranged the if block to avoid duplicating the warning about invalid servers.

## Motivation and Context 
I have a setup that uses an unbound container to provide DoT. I'm trying to integrate this composition as a stack in a swarm and it would greatly simplify the configuration if docker service names could be used. It would also make the behavior of this image more idiomatic. 

## How Has This Been Tested?
1. `ARCH=amd64 ./gh-actions-test.sh` 25 passed, 11 skipped
3. Using the docker-compose.yml  (below) with `PIHOLE_DNS_` set to "unbound;1.2.3.4" I started the composition.
4. I logged in to the pi-hole admin UI and verified the IP of the unbound container was set as the first custom server and 1.2.3.4 was the second. Then I removed the etc-* directories created in the working directory.
5. I repeated this with just IPv4 addresses.
6. I repeated this with just IPv6 addresses. The parts related to my change worked, but this produced an unrelated pihole-FTL error that exists in `pihole/pihole:latest` as well.
7. I repeated this with just the docker link.
Except for the IPv6 only test the specified IPs or IPs of specified service names were shown as the custom DNS servers and no errors were apparent in the logs.

## Types of changes

## Checklist:
- [x] My code follows the code style of this project. (I tried to be consistent with the surrounding code)

```
version: "2"

services:
  pihole:
    container_name: pihole
    image: pihole:support-docker-links-for-PIHOLE_DNS_dev-amd64-buster
    ports:
      - "53:53/tcp"
      - "53:53/udp"
      #- "67:67/udp"  # Required for DHCP
      - "80:80/tcp"
    environment:
      TZ: "America/New_York"
      PIHOLE_DNS_: "unbound;1.2.3.4"
    volumes:
      - "./etc-pihole:/etc/pihole"
      - "./etc-dnsmasq.d:/etc/dnsmasq.d"
    #cap_add:  # Required for DHCP
    #  - NET_ADMIN
    restart: unless-stopped

  unbound:
    image: haxwithaxe/unbound-dot:latest
```

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -127,9 +127,22 @@ if [ -n "${PIHOLE_DNS_}" ]; then
           change_setting "PIHOLE_DNS_$count" "$i"
           ((count=count+1))
           ((valid_entries=valid_entries+1))
-        else
-          echo "Invalid IP detected in PIHOLE_DNS_: ${i}"
+          continue
         fi
+        if [ -n "$(dig +short $i)" ]; then
+          # If the "address" is a domain (for example a docker link) then try to resolve it and add 
+          # the result as a DNS server in setupVars.conf.
+          resolved_ip="$(dig +short $i | head -n 1)"
+          echo "Resolved ${i} from PIHOLE_DNS_ as: ${resolved_ip}"
+          if valid_ip "$resolved_ip" || valid_ip6 "$resolved_ip" ; then
+            change_setting "PIHOLE_DNS_$count" "$resolved_ip"
+            ((count=count+1))
+            ((valid_entries=valid_entries+1))
+            continue
+          fi
+        fi
+        # If the above tests fail then this is an invalid DNS server
+        echo "Invalid IP detected in PIHOLE_DNS_: ${i}"
     done
 
     if [ $valid_entries -eq 0 ]; then
--- a/start.sh
+++ b/start.sh
@@ -129,10 +129,13 @@ if [ -n "${PIHOLE_DNS_}" ]; then
           ((valid_entries=valid_entries+1))
           continue
         fi
-        if [ -n "$(dig +short $i)" ]; then
+        if [ -n "$(dig +short ${i//#*/})" ]; then
           # If the "address" is a domain (for example a docker link) then try to resolve it and add 
           # the result as a DNS server in setupVars.conf.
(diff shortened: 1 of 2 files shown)


start.sh
```diff
@@ -127,9 +127,22 @@ if [ -n "${PIHOLE_DNS_}" ]; then
...
+          echo "Resolved ${i} from PIHOLE_DNS_ as: ${resolved_ip}"
+          if valid_ip "$resolved_ip" || valid_ip6 "$resolved_ip" ; then
+            change_setting "PIHOLE_DNS_$count" "$resolved_ip"
+            ((count=count+1))
+            ((valid_entries=valid_entries+1))
+            continue
+          fi
+        fi
+        # If the above tests fail then this is an invalid DNS server
+        echo "Invalid IP detected in PIHOLE_DNS_: ${i}"
```
PromoFaux, line 145:
```suggestion
        echo "Invalid entry detected in PIHOLE_DNS_: ${i}"
```

PromoFaux: While you are under the hood changing this, please could you change it inside the `if [ $valid_entries -eq 0 ]` block? thanks!

PromoFaux:
Thanks for the submission (apologies it's taken a while to review)

Can you please address the DCO sign-off requirement to satisfy the bot? thanks! https://github.com/pi-hole/docker-pi-hole/pull/1014/checks?check_run_id=5485296936

start.sh
haxwithaxe: Fixed both in a new commit.

haxwithaxe:
My last commit is broken :( Fixing now.

PromoFaux (approved):
Great stuff, thanks. Tested and working as described

networkException:
Thanks a lot for implementing this! I can finally decomission my patched start.sh then 😅

PR #1022, opened 2022-03-24
Test fix for moby security update.

dschaper:
Change the permissions to `permitted` instead of `inherited`.

See https://github.com/moby/moby/issues/43420#issuecomment-1077870013


PromoFaux (commented):
We can but try, approved pending test run

PR #1025, opened 2022-03-26
Use permitted instead of inherited caps.

dschaper:
The additional DAC_OVERRIDE permission should fix the docker engine security patch.


benleb:
can confirm this fixed the issue for me 👍 thanks @dschaper

dschaper:
Testing shows that `p` is not sufficient to create a working binary.

PR #1030, opened 2022-03-28
Docker 20.10.14 binary capabilites update.

dschaper:
Potential fix for the capabilities issue for docker engine 20.10.14.

Issue is from https://github.com/moby/moby/issues/43420#issuecomment-1077870013 and a security fix.

Due to a bug in the https://bugzilla.redhat.com/show_bug.cgi?id=1950187 library that wipes ambient permission when the effective user is changed, we have to set the capabilities in two steps. 

Set the effective and permitted to the binary via `setcap` to allow for ambient usage of those caps.

Start the binary with `capsh` and apply the capabilities for ambient there.

### DIFF
--- a/bash_functions.sh
+++ b/bash_functions.sh
@@ -3,7 +3,7 @@
 . /opt/pihole/webpage.sh
 
 fix_capabilities() {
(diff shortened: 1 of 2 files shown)


PromoFaux (approved):
LGTM, lets get this in dev and see what happens...

PR #1058, opened 2022-04-04
Replace one more instance of `touch` with `install /dev/null`

PromoFaux:
This one was missed, noticed issues when running current images on `buster` (without updated `libseccomp`), this PR brings the usage in line with the changes that were made to the FTL S6 service script

### DIFF
--- a/bash_functions.sh
+++ b/bash_functions.sh
@@ -47,7 +47,7 @@ prepare_configs() {
     installConfigs
    
     if [ ! -f "${setupVars}" ]; then
-        install /dev/null "${setupVars}"
+        install -m 644 /dev/null "${setupVars}"
         echo "Creating empty ${setupVars} file."
     fi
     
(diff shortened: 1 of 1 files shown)


bash_functions.sh
```diff
@@ -47,7 +47,7 @@ prepare_configs() {
     installConfigs
    
     if [ ! -f "${setupVars}" ]; then
-        touch "${setupVars}"
+        install /dev/null "${setupVars}"
```
yubiuser, line 50:
```suggestion
        install -m 644/dev/null "${setupVars}"
```
?

rdwebdesign: Or `install -m 644 -o pihole -g pihole /dev/null "${setupVars}"`?

PromoFaux: Maaaaybe. I was only going by it having a plain `touch` with no mode change before. Do you think we should set the mode here?

yubiuser (approved):

PR #1084, opened 2022-05-03
Update default debian to bullseye. Update README to describe docker tags.

lightswitch05:
## Description

It looks like Bullseye is already the default (only) debian version we are using - so bump the other defaults to bullseye as well. 

Also, update the README to better match the current docker tagging pattern.

## Motivation and Context

Mostly just some cleanup as I catch back up on this project

## How Has This Been Tested?

No testing other then Github Actions - the defaults shouldn't really change anything functional since we're already building bullseye images.

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,23 +222,16 @@ Users of older Ubuntu releases (circa 17.04) will need to disable dnsmasq.
 
 ## Docker tags and versioning
 
-The primary docker tags / versions are explained in the following table.  [Click here to see the full list of tags](https://store.docker.com/community/images/pihole/pihole/tags), I also try to tag with the specific version of Pi-hole Core for version archival purposes, the web version that comes with the core releases should be in the [GitHub Release notes](https://github.com/pi-hole/docker-pi-hole/releases).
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -222,23 +222,16 @@ Users of older Ubuntu releases (circa 17.04) will need to disable dnsmasq.
...
+The primary docker tags are explained in the following table.  [Click here to see the full list of tags](https://store.docker.com/community/images/pihole/pihole/tags). See [GitHub Release notes](https://github.com/pi-hole/docker-pi-hole/releases) to see the specific version of Pi-hole Core, Web, and FTL included in the release.
+
+| tag                 | description                                                                                                                                
+|---------------------|--------------------------------------------------------------------------------------------------------------------------------------------|
+| `latest`            | Always latest release                                                                                                                      |
+| `2022.04`           | Date-based release that can receive bugfix updates                                                                                         |
+| `2022.04.1`         | A specific image that will not receive updates                                                                                             |
```
PromoFaux, line 231: As it stands, a `2022.04.1` will not also tag `2022.04` again. Maybe it should (I think most people just use `latest` in any case

But thanks for looking over this section - I completely didn't see it before!

dschaper: I'd like it if it did. That's how a few images I use do it. So you can use `2022.04` and always have the latest patches/fixes.

PR #1101, opened 2022-05-23
Allow pihole checkout on any image rather than just dev/nightly image…

PromoFaux:
## Description

Allows `pihole checkout` to be run, even on versioned/`:latest` tags, in my opinion there isn't a real need to be supressing this functionality in a docker container - it is still useful for troubleshooting

## Motivation and Context

https://github.com/pi-hole/FTL/issues/1348#issuecomment-1133841419

## Types of changes

## Checklist:


lightswitch05 (approved):
LGTM

PR #1105, opened 2022-05-26
Bake latest release (at time of build) of PADD into the container

PromoFaux:
## Description

Simple really, adds https://github.com/pi-hole/padd into the container

## Motivation and Context

We started an internal discussion, and then I did this.

## How Has This Been Tested?

`docker exec -it pihole padd`

padd

## Types of changes

## Checklist:


thomasmerz:
🎗️ Reminder for me: update README in https://github.com/pi-hole/PADD/pull/208 again when this PR is finished.

PromoFaux:
I've not forgotten about this, just not 100% how I want to achieve the end result. 

Another thought that has crossed my mind is adding the download of PADD into the container startup, so that it is _always_ up to date

rdwebdesign:
> Another thought that has crossed my mind is adding the download of PADD into the container startup, so that it is _always_ up to date

I think this is a very good idea. It will avoid the need of updating the docker image each time PADD is updated.

rdwebdesign:
Do we need the whole repo?
Maybe it would be better to copy just the script using `wget`. What do you think?

Dockerfile
```diff
@@ -25,6 +25,10 @@ ENV PHP_ERROR_LOG /var/log/lighttpd/error.log
 COPY ./start.sh /
 COPY ./bash_functions.sh /
 
+# Add PADD to the container, too.
+ADD https://raw.githubusercontent.com/pi-hole/PADD/master/padd.sh /padd.sh
```
rdwebdesign, line 29: I think we could use `https://install.padd.sh` here, since we already  have a [PR to change the URL](https://github.com/pi-hole/PADD/pull/213)

rdwebdesign (changes requested):

Dockerfile
PromoFaux:
```suggestion
ADD https://install.padd.sh /padd.sh
```

PR #1117, opened 2022-06-17
set NAMESERVER on startup if PH_NAMESERVER is set

mhus:
Set the name server IP if defined in PH_NAMESERVER environment

## Description

Not in every case it is possible to set the DNS at start of the docker image. In my case Im using snap and I can't set the dns on startup. In this case pinhole do not start up because name server entry is 127.0.0.1 by default.

On startup add --env PH_NAMESERVER=<ip>

## Motivation and Context

At the moment for every new version of PI Hole Container I need to edit the start file and set the name server manually. This change will fix this.

## How Has This Been Tested?

Tetsed on my qnap container station.

## Types of changes

## Checklist:

### DIFF
--- a/s6/debian-root/etc/cont-init.d/20-start.sh
+++ b/s6/debian-root/etc/cont-init.d/20-start.sh
@@ -7,6 +7,10 @@ if [ "${PH_VERBOSE:-0}" -gt 0 ] ; then
     bashCmd='bash -e -x'
 fi
 
+if [ "${PH_NAMESERVER:-0}" -gt 0 ] ; then
+    echo nameserver ${PH_NAMESERVER} > /etc/resolv.conf
+fi
+
 $bashCmd /start.sh
 # Gotta go fast, no time for gravity
 if [ -n "$PYTEST" ]; then


PromoFaux:
Thanks for the contribution! Hmm, I'm  unsure about this. How are you defining your container? (compose file? run command?)

https://docs.docker.com/config/containers/container-networking/#dns-services

What is different about snap (not quite sure I understand how you're using it here!) that it cannot set the DNS in the container configuration as above?

PromoFaux:
I understand where you're coming from, but I don't feel like this is the correct way to solve the issue and probably falls under the "edge case" category

The issue is, there is a proper way to do it - it's just that Container Station on QNAP doesn't support it, seemingly - I've not ever used it, so my experience here is based only on a few quick google searches.

 Apparently you can [install and use Portainer](https://medium.com/digitalfrontiers/easily-set-up-portainer-ce-on-a-qnap-nas-5d82b2390854) (or maybe [this one?](https://www.qnapclub.eu/en/qpkg/466)) as an alternative (which I use myself in a couple of places) if you really need a GUI option to create your containers - else, does QNAP not allow you to create a docker container on the command line? This should give you the ability to pass the `--dns`  option

mhus:
No its not really good working on QNAP. I will create a feature request in QNAP for it. But this will take a while to be done. In the meantime I can extend the container and add the fix for myself. That's also ok for me.

Issue #1137, opened 2022-07-10
macvendor.db is not created


moritzbeck01:
This is a: **Bug**  

## Details

The `macvendor.db` file is missing from the `/etc/pihole` directory in the container and is not automatically created on container startup or (re-)creation.

## Related Issues
- [x] I have searched this repository/Pi-hole forums for existing issues and pull requests that look similar 

[Reddit - macvendor.db missing](https://www.reddit.com/r/pihole/comments/co48l1/macvendordb_missing/)
[GitHub - Recent PR and how I noticed this issue](https://github.com/pi-hole/FTL/issues/1351) 

## How to reproduce the issue 

1. Environment data
  * Operating System: **Raspberry Pi OS 64-bit Lite** 
  * Hardware: **Raspberry Pi 4 Model B 4GB RAM** 
  * Kernel Architecture: **aarch64** 
  * Docker Install Info and version: 
    - Software source: **official Docker convenience script** 
    - Supplimentary Software: **nothing that interfaces Pi-hole directly, a nginx reverse proxy for the web interface** 
  * Hardware architecture: **ARMv8**

2. docker-compose.yml contents, docker run shell command, or paste a screenshot of any UI based configuration of containers here

docker run script:
(I use the same template for all my containers and then just change the variables at the top - sorry for formatting)
```
image=pihole/pihole:latest
name=pihole
#number=10
volume1=pihole
volume2=dnsmasq.d
docker pull $image
docker stop $name
docker rm -f $name
docker run \
        --detach \
        --dns 8.8.8.8 \
        --dns 8.8.4.4 \
        --env-file $name.env \
        --hostname $name \
        --ip 10.0.10.$number \
        --mac-address 02:42:10:00:10:$number \
        --name $name \
        --network macvlan \
        --restart unless-stopped \
        --volume $(pwd)/$name/$volume1:/etc/$volume1 \
        --volume $(pwd)/$name/$volume2:/etc/$volume2 \
        --cap-add NET_ADMIN \
        $image
```

3. any additional info to help reproduce

docker environment file:
(sorry for formatting again)
```
TZ=Europe/Berlin
WEBPASSWORD=censored
FTLCONF_REPLY_ADDR4=10.0.10.10
ADMIN_EMAIL=censored
PIHOLE_DNS_=8.8.8.8;8.8.4.4
DNSSEC=true
DNS_BOGUS_PRIV=true
DNS_FQDN_REQUIRED=true
REV_SERVER=false
[...]

PR #1138, opened 2022-07-10
download macvendor.db on container start if it is missing

PromoFaux:
Addresses https://github.com/pi-hole/docker-pi-hole/issues/1137.

In case of a user mounting `/etc/pihole` to their host, this redownloads the `macvendor.db` file used by the network table.

Ordinarily this is downloaded during the original install process (and, in fact, is present in the image), but obviously overriding that directory will clear it out. There is no other mechanism in the base install that re-acquires a missing `macvendor.db`, as this is not ordinarily a problem on a bare metal install, and the file doesn't change an awful lot

### DIFF
--- a/bash_functions.sh
+++ b/bash_functions.sh
@@ -69,6 +69,12 @@ prepare_configs() {
     if [[ -e "${setupVars}" ]]; then
         cp -f "${setupVars}" "${setupVars}.update.bak"
     fi
+
+    #If the user has volume mounted /etc/pihole, then macvendor.db may be missing. See: https://github.com/pi-hole/docker-pi-hole/issues/1137
+    if [[ ! -f "/etc/pihole/macvendor.db" ]]; then
+        echo "Downloading missing macvendor.db"
+        curl -sSL "https://ftl.pi-hole.net/macvendor.db" -o "/etc/pihole/macvendor.db" || true
+    fi
 }
 
 validate_env() {


billygr:
Hi,

The proposed solution just creates a symlink macvendor.db -> /macvendor.db ( in my setup (when upgraded from 
[2022.07.1](https://github.com/pi-hole/docker-pi-hole/releases/tag/2022.07.1) to [2022.08](https://github.com/pi-hole/docker-pi-hole/releases/tag/2022.08)

```
    # Volumes store your data between container upgrades
    volumes:
      - './etc-pihole:/etc/pihole'
      - './etc-dnsmasq.d:/etc/dnsmasq.d'
```
I have remove it manually and started stopped the container same issue.

The fix saves the file in the root (/) of the container (and i don't think that the /etc/pi-hole is mounted at the stage we download the file)

Expectations is that the file will reside on the mounted etc-pihole directory (which i may be also wrong)

The symlink is working as expected when i run a shell inside the container

PromoFaux:
> The proposed solution just creates a symlink

This is correct. A fresh `macvendor.db` is actually downloaded as part of the container creation process (before it is uploaded to docker hub) when Pi-hole is installed. At the end of the install process, that `macvendor.db` is then moved out to `/` so that any outdated copy of this file can be "updated" by creating a symlink. (thinking on it, not sure why we don't just copy it tbh)

https://github.com/pi-hole/docker-pi-hole/blob/8619de00315e0cfb03d3d166babc2d246313f980/src/scripts/install.sh#L86-L90

On container startup:

https://github.com/pi-hole/docker-pi-hole/blob/8619de00315e0cfb03d3d166babc2d246313f980/src/scripts/bash_functions.sh#L106-L110

billygr:
Thank you @PromoFaux that explains it

PR #1141, opened 2022-07-11
Huge tidy up/prune/refactor of the code

PromoFaux:
A lot of changes here - but this has been a long time coming, and is desperately needed. Tested functionality as I went along, everything appears to be in order.

Summary of changes:

- Moves most code in start.sh into functions in bash_functions.sh
- Removes code from 20-start.sh and moves it inside start.sh
- Removes unused variables and code, making it easier to find our way around
- Removes dependency on webpage.sh, which had to be modified in order to source it properly (thus breaking pihole checkout), use utils.sh instead
- Replaces all uses of ServerIP/v6 with new FTLCONF_REPLY_ADDR4/6 variables
- No need to symlink echo to whiptail any more (probably hasn't been needed for a while)
- removes huge list of exported env vars at the top of start.sh - no longer appear to be needed
- rename some functions in bash_functions to more logical names, group their usages in start.sh
- adjust/fix tests to match changes
- remove some dead tests
- Upgrade to S6-overlay v3


s6/debian-root/etc/s6-overlay/s6-rc.d/1-startup/up
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/execlineb
```
rdwebdesign, line 1: I'm not sure, but I think this should be:

```suggestion
#!/command/execlineb
```

Issue #1149, opened 2022-07-20
s6v3: gravity update at boot time timeout on rpi3.


edgd1er:
the branch webpassword-env, in its current state does not start: (gravity update timeout, laptop i5 ).
changing stage to 1, allow the container to continue without stopping all services.
setting SKIPGRAVITYONBOOT=1, disable gravity update on boot as expected, there is no timeout.

```
docker-pi-hole-pihole-1  |   [i] Target: https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts
  [✓] Status: Retrieval successful
docker-pi-hole-pihole-1  | s6-rc: fatal: timed out
docker-pi-hole-pihole-1  | /run/s6/basedir/scripts/rc.init: warning: s6-rc failed to properly bring all the services up! Check your logs (in /run/uncaught-logs/current if you have in-container logging) for more information.
docker-pi-hole-pihole-1  | prog: fatal: stopping the container.
docker-pi-hole-pihole-1  | s6-sudoc: fatal: unable to get exit status from server: Operation timed out
docker-pi-hole-pihole-1  | s6-rc: info: service legacy-cont-init: stopping
docker-pi-hole-pihole-1  | s6-rc: info: service lighttpd: stopping
docker-pi-hole-pihole-1  | s6-rc: info: service cron: stopping
docker-pi-hole-pihole-1  | Stopping cron
docker-pi-hole-pihole-1  | s6-rc: info: service cron successfully stopped
docker-pi-hole-pihole-1  | Stopping lighttpd
docker-pi-hole-pihole-1  | s6-rc: info: service legacy-cont-init successfully stopped
docker-pi-hole-pihole-1  | s6-rc: info: service fix-attrs: stopping
docker-pi-hole-pihole-1  | s6-rc: info: service fix-attrs successfully stopped
docker-pi-hole-pihole-1  | s6-rc: info: service lighttpd successfully stopped
docker-pi-hole-pihole-1  | s6-rc: info: service pihole-FTL: stopping
docker-pi-hole-pihole-1  | Stopping pihole-FTL
docker-pi-hole-pihole-1  | s6-rc: info: service pihole-FTL successfully stopped
docker-pi-hole-pihole-1  | s6-rc: info: service 1-startup: stopping
docker-pi-hole-pihole-1  | s6-rc: info: service 1-startup successfully stopped
docker-pi-hole-pihole-1  | s6-rc: info: service s6rc-oneshot-runner: stopping
[...]

PR #1152, opened 2022-07-21
Run `_gravityonboot` oneshot in background

PromoFaux:
Attempts to solve https://github.com/pi-hole/docker-pi-hole/issues/1149 by adding an infinite value to the timeout of this script.


PromoFaux:
@edgd1er - Please try this on your Pi

PromoFaux:
It is not this. Still times out:

image

PromoFaux:
Should work now...

edgd1er:
@PromoFaux , a bit late, nonetheless I confirm the fix is ok on my rpi3, no more timeout while updating the database. Thanks.

PR #1235, opened 2022-11-09
Added Dockerfile for local usage.

SamTV12345:
This pr solves multiple issues:
- As stated here [Discourse Pihole](https://discourse.pi-hole.net/t/how-to-build-a-recent-version-of-docker-pihole-locally/59074/3) a local Dockerfile is currently not present. This local dockerfile can be executed by every user with ``docker build -t local_pihole -f Dockerfile_local .``
- Decrease vulnerabilities from critical to medium, decrease number of vulnerabilities by 135.

The current build with the latest debian image has 178 known vulnerabilities:
- 11 critical
- 32 high
- 43 middle
- 92 low 
image

The local image is based on Ubuntu and has 43 vulnerabilties
- 8 middle
- 35 low
image

### DIFF
--- a/src/Dockerfile_local
+++ b/src/Dockerfile_local
@@ -1,13 +1,25 @@
 ARG PIHOLE_BASE
-FROM "${PIHOLE_BASE:-ghcr.io/pi-hole/docker-pi-hole-base:bullseye-slim}"
+FROM ubuntu:latest
+
+ENV TZ=Europe/Berlin
+ENV WEBPASSWORD=changeme
+
+
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+
+RUN apt update -y && apt upgrade -y && apt install php curl -y
 
 ARG PIHOLE_DOCKER_TAG
 RUN echo "${PIHOLE_DOCKER_TAG}" > /pihole.docker.tag
 
 ENTRYPOINT [ "/s6-init" ]
 
+ # etc and usr are now in /
 COPY s6/debian-root /
+# file is in /usr/local/bin/service
 COPY s6/service /usr/local/bin/service
+
+WORKDIR /usr/local/bin
 
 RUN bash -ex install.sh 2>&1 && \
     rm -rf /var/cache/apt/archives /var/lib/apt/lists/*


SamTV12345:
Sorry for force pushing had to add my signature and verification via gpg to the commit.

PromoFaux:
To quote from your discourse post:

>If I want to run the Dockerfile in src I get errors that the path to the scripts aren't valid.


This is what I run for building a local image when testing:

```
docker build ./src/ --build-arg PIHOLE_DOCKER_TAG=dev --tag adamtest --no-cache
```

At the moment, I'm not really seeing the benefit to this PR - The current base image is based on debian bullseye: https://github.com/pi-hole/docker-base-images/blob/master/docker-pi-hole-base/Dockerfile

If Ubuntu is a a better fit, then maybe there would be the best place to change the base image - but at the moment I don't know what the advantage is.  What tool are you using to scan for vulnerabilities?

SamTV12345:
> ```
> docker build ./src/ --build-arg PIHOLE_DOCKER_TAG=dev --tag adamtest --no-cache
> ```
Thank you for posting your docker build command. Now it makes sense why the path were incorrect because I ran it directly in src. 


> At the moment, I'm not really seeing the benefit to this PR - The current base image is based on debian bullseye: https://github.com/pi-hole/docker-base-images/blob/master/docker-pi-hole-base/Dockerfile

I tinkered around with the base Debian Pihole image and also ran ``apt update && apt upgrade`` but the number and severity of the vulnerabilities stayed the same. Seems like Debian is currently not that secure.

> If Ubuntu is a a better fit, then maybe there would be the best place to change the base image - but at the moment I don't know what the advantage is. What tool are you using to scan for vulnerabilities?

That would be great. I used [harbor](https://goharbor.io/) which uses [aquasec trivy](https://hub.docker.com/r/aquasec/trivy) to scan for vulnerabilities  You can test it locally with ```docker run aquasec/trivy image pihole/pihole``` . A good addition would be to rebuilt the current image every month unless there is a new release which already triggers the build pipeline. That way the base image is at most 1 month old.

dschaper:
Can you help me understand what the attack vector is? I don't see how this applies to a container?

SamTV12345:
A container is not really that safe. Yes I can't break out of the container but I can still reach everything that is inside my local network. So with a buffer overflow I could execute arbitrary code inside the pod/docker container and because an intruder would have one of the most important services in my homelab under control this is a problem. He would not need to do something like DNS cache poising because he essentially owns the pihole dns cache.

E.g. when you want to visit gitlab.com an attacker could simply redirect you to gitIab.com with a local dns entry. One time with an I and one time with a little L in the middle and because the container runs inside as root the intruder can always add other Linux repositories, install new applications and start other attack vectors.

To summarize: Keeping your containers up to date and without vulnerabilities in the high and critical sector is always a good practice. If you have a server for your pihole development I recommend installing Harbor and schedule scans for the pihole image. That way you can check if the images are still safe to use.

PromoFaux:
That all said - thank you for the contribution - it is appreciated, even if we don't (currently) agree. I still don't really see the benefits of switching to a Ubuntu base - 43 vulnerabilities is not 0 vulnerabilities, so the container would be just as hypothetically vulnerable as it may or may not be now. 

(As always we encourage responsible _specific_ vulnerability disclosure via disclosure@pi-hole.net - please take this as a challenge, we are here to learn)

Also, excuse the "whataboutism" here for a sec, but I'd argue that a lot of the following are also images one wouldn't want to be vulnerable - and yet....

```
nginx (debian 11.5)
===================
Total: 141 (UNKNOWN: 0, LOW: 93, MEDIUM: 25, HIGH: 20, CRITICAL: 3)

python:latest (debian 11.5)
===========================
Total: 1254 (UNKNOWN: 3, LOW: 595, MEDIUM: 278, HIGH: 362, CRITICAL: 16)

postgres:latest (debian 11.5)
=============================
Total: 111 (UNKNOWN: 0, LOW: 88, MEDIUM: 7, HIGH: 14, CRITICAL: 2)

redis:latest (debian 11.5)
==========================
Total: 77 (UNKNOWN: 0, LOW: 61, MEDIUM: 4, HIGH: 10, CRITICAL: 2)

httpd:latest (debian 11.5)
==========================
Total: 99 (UNKNOWN: 0, LOW: 75, MEDIUM: 8, HIGH: 14, CRITICAL: 2)

node:latest (debian 11.5)
=========================
Total: 1211 (UNKNOWN: 3, LOW: 572, MEDIUM: 272, HIGH: 350, CRITICAL: 14)

php:8.2.0RC6-fpm-bullseye (debian 11.5)
=======================================
Total: 349 (UNKNOWN: 3, LOW: 253, MEDIUM: 49, HIGH: 42, CRITICAL: 2)
```

To be honest, if the debian base is good enough for those projects (I literally just went down the top official docker repositories), then personally it's good enough for me at this stage. 

If there was an overwhelming benefit to ubuntu, I would consider it more.

Question: Have you read over all of the reported vulns spat out by the analytics tool?

dschaper:
> E.g. when you want to visit gitlab.com an attacker could simply redirect you to gitIab.com with a local dns entry

No, you can't. You could return a different IP address for a domain lookup but that's about it. You'd have to do a whole lot of TLS trickery in addition to do anything nefarious.

SamTV12345:
> > E.g. when you want to visit gitlab.com an attacker could simply redirect you to gitIab.com with a local dns entry
> 
> No, you can't. You could return a different IP address for a domain lookup but that's about it. You'd have to do a whole lot of TLS trickery in addition to do anything nefarious.

There must not be a lot of TLS trickery. I just reserve the a domain that is close in terms of spelling to the source domain, adapt the dns entry with my custom IP address for the real source domain and redirect when someone accesses the

PR #1238, opened 2022-11-17
Fix command example to be docker-compose

philverghese:
The example says `docker compose` which I don't think is correct

Signed-off-by: Phil Verghese <philverghese@users.noreply.github.com>

## Description
The command given in the quick start example uses `docker` instead of `docker-compose`

## Motivation and Context
Give a command line that works

## How Has This Been Tested?
Ran locally

## Types of changes

## Checklist:

- [] My code follows the code style of this project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ services:
       - NET_ADMIN # Required if you are using Pi-hole as your DHCP server, else not needed
     restart: unless-stopped
 ```
-2. Run `docker compose -f docker-compose.yml up -d` to build and start pi-hole
+2. Run `docker-compose -f docker-compose.yml up -d` to build and start pi-hole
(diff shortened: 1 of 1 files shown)


PromoFaux:
I kind of thought this might happen when I accepted the patch that changed it to the new style command. 

My reason for accepting it was that If you follow the link in the docs (line 28) for how to install compose, you would actually end up with a system using the new syntax.

PromoFaux:
https://docs.docker.com/compose/install/

Note scenario three is the only one that would result in `docker-compose` Vs `docker compose`, and that scenario is explicitly marked as unsupported

philverghese:
I think the problem is I'm running Docker on a Synology NAS, which apparently uses the third (now unsupported) scenario.

For me, I need to run the command as `docker-compose` which is preinstalled as part of the Synology Docker package.

PromoFaux:
@philverghese was just reminded of this by #1264 and noted that you can install `docker compose` via scenario two.

https://docs.docker.com/compose/install/linux/#install-the-plugin-manually

Here is the result of doing so on my own Syno running DSM 7:

image

Issue #1249, opened 2022-11-30
Date based tag like "2022.04" is absolute and does not receive bugfix updates


quotidian-ennui:
It appears that the ghcr.io images don't match the desired state that the documentation suggests.

This differs from what's in the documentation about versioning 

> 2022.04 : Date-based release that can receive bugfix updates

I _believe_ that it might be as easy as having in your github action that gets emits the docker metadata

https://github.com/pi-hole/docker-pi-hole/blob/a787e29aad4604a42eee6f553fd6df6f08461b1e/.github/workflows/test-and-build.yaml#L56-L59

```
        tags: |
          type=semver,pattern={{version}}
          type=semver,pattern={{major}}.{{minor}}
          type=schedule 
          type=ref,event=branch,enable=${{ github.event_name != 'schedule' }} 
          type=ref,event=tag 
```

This should generate the tags `2022.11.2 2022.11` against the same image.

## Details

Check the build date and digest for the images 2022.11 && 2022.11.2 in the github package registry (the dates differ).

This also appears to be the case for the associated images from hub.docker.com

(sorry I didn't fill in the rest of the template).

PromoFaux:
It might just need the documentation updating.

In all reality the version number is only to differentiate from the previous build. The number after the month could just as easily be the day of the month - it doesn't really follow semantic versioning, and so 2022.01.1 does not nessacerily constitute re-tagging 2022.01

There is no real sensible way to do semantic versioning with this image - as it is an omni-container with three distinct packages, each with their own versions, inside. 

General rule of thumb is first build of the month gets `yyyy.mm` any additional builds that month get an incremented number at the end.

PromoFaux:
And just to further add - I would not want to retag any known working tag with something thay could potentially be broken (it happens, we aren't perfect). As it is right now, people always have the option to roll back to the previous known working image.

rdwebdesign:
I think the highlighted text creates a confusion: 
_"Date-based release **that can receive bugfix updates**"_.

If an update is released, there will be a new label tag. I would be wrong to change the contents of a previous image.

PromoFaux:
Absolutely - that's the part that needs rewording. Feel free to PR something if you think of anything - I am currently on the sofa and cannot be bothered to move to the computer!

dschaper:
I think we discussed this before as "floating tags".

[...]

PR #1250, opened 2022-11-30
Update README to explain the version numbers slightly more accurately

PromoFaux:
See: https://github.com/pi-hole/docker-pi-hole/issues/1249

The documentation as it is is confusing, the container does not follow any kind of semver rule, and also previous tags are not re-tagged to include "bugfixes".

New builds are just that... new builds. We give them some kind of version number for easy switching back and forth between them


PromoFaux:
a

PromoFaux:
a

PromoFaux:
a

mwolter805:
Hello, It appears there may be a bit more ambiguity regarding versioning as there will never be a 2022.12.0 version. In order to programmatically decipher versions, it might be useful to say something like the following:

Container versioning does not follow [semver](https://semver.org), and can follow two different formats. 

Version format A has two octets, with the first octet representing the year and the second octet representing the month of the release. For example 2022.12

Version format B has three octets, with the first two octets following format A and a third octet representing the update number for the particular month. For example 2022.11.2

Note: the third octet of format B will only be >= 1. For example, 2022.12.0 is invalid.

PR #1264, opened 2022-12-28
Changed the build and compose command syntax in README.md to be functional

zleyyij:
eral summary of your changes in the Title above -->

## Description

The readme had `docker compose -f docker-compose.yml up -d`, which references the wrong binary. It should be updated to reference `docker-compose`, a different utility than `docker`, and the correct binary to complete the operation at hand.

## Motivation and Context

This change was added to make the example command for building and deploying functional.  Prior to this change the command would fail, because the syntax was meant for the `docker-compose` binary and does not work with the `docker` binary
## How Has This Been Tested?
This was tested using the preview feature and by running the command in a Debian test environment with `docker` and `docker-compose installed

## Types of changes

## Checklist:

- [x ] I have updated the documentation accordingly.

Signed-off-by: zleyyij <75810274+zleyyij@users.noreply.github.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ services:
       - NET_ADMIN # Required if you are using Pi-hole as your DHCP server, else not needed
     restart: unless-stopped
 ```
-2. Run `docker compose -f docker-compose.yml up -d` to build and start pi-hole
+2. Run `docker-compose -f docker-compose.yml up -d` to build and start pi-hole
 3. Use the Pi-hole web UI to change the DNS settings *Interface listening behavior* to "Listen on all interfaces, permit all origins", if using Docker's default `bridge` network setting. (This can also be achieved by setting the environment variable `DNSMASQ_LISTENING` to `all`)
 
 [Here is an equivalent docker run script](https://github.com/pi-hole/docker-pi-hole/blob/master/examples/docker_run.sh).


edgd1er:
According to the docker [documentation](https://docs.docker.com/engine/reference/commandline/compose/): `docker compose`is the command now included in the docker project. 

According to the docker/compose documentation, docker-compose is the python v1 program. docker compose is the [Go v2 program](https://github.com/docker/compose/). docker-compose is an external program.
```
For a smooth transition from legacy docker-compose 1.xx, please consider installing [compose-switch](https://github.com/docker/compose-switch) to translate docker-compose ... commands into Compose V2's docker compose .... . Also check V2's --compatibility flag.
```
why should docker compose be the wrong binary ?
is the README example not compatible with docker compose command ?
what is the version of the installed compose ?

DL6ER:
@edgd1er is right, any recent enough docker installation will not know about the `docker-compose` tool and provide `compose` as a `docker` command instead. However, also due to the partially extremely outdated `docker` installations out there, I do see some value in adding a statement like: "If `docker compose` doesn't work on your system, try using `docker-compose` instead."

PromoFaux:
Bullet point 1 in the Quick start section of the readme contains a link to : https://docs.docker.com/compose/install/

From that link, it notes that `docker-compose` is now an unsupported syntax - though as it was pointed out in the [previous pull request with the same intention](https://github.com/pi-hole/docker-pi-hole/pull/1238), Synology is running an older version of Docker ([20.10.3](https://www.synology.com/en-uk/releaseNote/Docker) according to mine at time of writing)

That said, I have just installed it [manually](https://docs.docker.com/compose/install/linux/#install-the-plugin-manually) via the instructions in the provided install link...

image

So:

> I do see some value in adding a statement like: "If docker compose doesn't work on your system, try using docker-compose instead."

I don't think we need to add this note - but maybe something that makes it more obvious that clicking the link will lead you to a place that lets you know how to install `docker compose` on your system.

zleyyij:
This was on a brand new 64 bit Raspbian installation, updated with `sudo apt update` and `sudo apt upgrade`. Docker was installed via `apt`.
image
I did not know about the newer version of docker, however I feel that because the version of docker in the Debian repos does still not support compose, it should at least be noted.

PromoFaux:
Ah, I've never installed docker directly through `apt` on raspbian, in fact on any system (aside from syno) I usually run the convenience script instead to install it:

https://docs.docker.com/engine/install/debian/#

This gets you the latest docker instead of relying on whatever Debian/raspberry pi OS ships

PR #1266, opened 2023-01-02
Readme change - Docker compose to docker-compose

jmilne22:
My friend was going through this today and was having issues running `docker-compose` unless I am mistaken the `docker compose` in the README should be `docker-compose`.

## Description
Changed `docker compose` to `docker-compose` in the readme

## Motivation and Context
Some users may run `docker compose` and get an error.

## Types of changes
My change requires a change to the documentation on the README.


jmilne22:
Nevermind, I did some reading and it turns out `docker compose` does exist, so I assume this is intentional.

PR #1339, opened 2023-03-23
Recreate pi-hole cronjob on container start

casperklein:
## Description

This PR recreates the pi-hole cronjob on container start.

## Motivation and Context

I stumbled about this:

```
# Pi-hole: Update the ad sources once a week on Sunday at a random time in the
#          early morning. Download any updates from the adlists
#          Squash output to log, then splat the log to stdout on error to allow for
#          standard crontab job error handling.
17 4   * * 7   root    PATH="$PATH:/usr/sbin:/usr/local/bin/" pihole updateGravity >/var/log/pihole/pihole_updateGravity.log || cat /var/log/pihole/pihole_updateGravity.log
```

In a regular pi-hole setup, a random time is generated to run the cron job. When running pi-hole in a container, the time is not random and everyone that uses the same image, will update gravity at the same time.

## How Has This Been Tested?

I checked the cronjob on container start. It has different times on each container start.

## Types of changes

## Checklist:

### DIFF
--- a/src/s6/debian-root/usr/local/bin/_startup.sh
+++ b/src/s6/debian-root/usr/local/bin/_startup.sh
@@ -43,6 +43,7 @@ setup_lighttpd_bind
 
 # Misc Setup
 # ===========================
+installCron
 setup_blocklists
 
 # FTL setup


PromoFaux (approved):
Thanks, this must have gone walkabout at some point in the past.

PR #1344, opened 2023-04-15
Trigger stale workflow on issue comments to remove stale label immediately

yubiuser:
**What does this PR aim to accomplish?:**

We use the `stale` action to mark issues stale after 30 days of inactivity. The actions runs only once a day to mark stale issues. If updates to issues occur (e.g. comments) the action removes the stale label. However, this happens so far only once a day - in the worst case leaving a "gap" of 23:59 between comments and label removal.

This PR adds a trigger to the stale workflow to run on `issue_comment`so that new comments should remove the label immediately.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -6,6 +6,9 @@ on:
   workflow_dispatch:
   issue_comment:
 
+env:
+  stale_label: stale
+
 jobs:
   stale_action:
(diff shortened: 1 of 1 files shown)


yubiuser:
Needs rebasing after https://github.com/pi-hole/docker-pi-hole/pull/1345 has been merged.

PromoFaux (approved):

PR #1352, opened 2023-05-11
Tweak commands

yubiuser:
## Description
Tweaks three commands
- detect the architecture using `arch`. Not all distros use `dpkg`
- `killall` requires the `-s|--signal` flag for any signal execpt `SIGTERM`
- only delete files that exist

## Motivation and Context
Makes the image more portable. Mountain is coming.

## Types of changes

### DIFF
--- a/src/s6/debian-root/usr/local/bin/install.sh
+++ b/src/s6/debian-root/usr/local/bin/install.sh
@@ -10,7 +10,7 @@ WEB_LOCAL_REPO=/var/www/html/admin
 setupVars=/etc/pihole/setupVars.conf
 
 detect_arch() {
-  DETECTED_ARCH=$(dpkg --print-architecture)
+  DETECTED_ARCH=$(arch)
   S6_ARCH=$DETECTED_ARCH
   case $DETECTED_ARCH in
   amd64)
@@ -94,9 +94,13 @@ mv /etc/pihole/macvendor.db /macvendor.db
 
 
 ## Remove the default lighttpd unconfigured config:
-rm /etc/lighttpd/conf-enabled/99-unconfigured.conf
+if [ -f /etc/lighttpd/conf-enabled/99-unconfigured.conf ]; then
+  rm /etc/lighttpd/conf-enabled/99-unconfigured.conf
+fi
 ## Remove the default lighttpd placeholder page for good measure
-rm /var/www/html/index.lighttpd.html
+if [ -f /var/www/html/index.lighttpd.html ]; then
+  rm /var/www/html/index.lighttpd.html
+fi
 ## Remove redundant directories created by the installer to reduce docker image size
 rm -rf /tmp/*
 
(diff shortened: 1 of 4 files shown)


src/s6/debian-root/usr/local/bin/install.sh
```diff
@@ -10,7 +10,7 @@ WEB_LOCAL_REPO=/var/www/html/admin
 setupVars=/etc/pihole/setupVars.conf
 
 detect_arch() {
-  DETECTED_ARCH=$(dpkg --print-architecture)
+  DETECTED_ARCH=$(arch)
```
rdwebdesign, line 13: This won't return the same values:

```console
pi@rdweb:~ $ dpkg --print-architecture
armhf

pi@rdweb:~ $ arch
armv7l
```

rdwebdesign (changes requested):

src/s6/debian-root/usr/local/bin/install.sh
yubiuser: You're right. I tested on two devices I have here and it even simplifies the logic

```
chrko@ThinkPad-X230:~$ arch
x86_64
chrko@ThinkPad-X230:~$ dpkg --print-architecture
amd64
```

```
nanopi@nanopi:~$ dpkg --print-architecture
arm64
nanopi@nanopi:~$ arch
aarch64
```

rdwebdesign: I think there are more possible architectures here.

Testing on a Raspberry Pi 4B running Bullseye I got this:

```console
pi@raspberrypi:~ $ dpkg --print-architecture
arm64

pi@raspberrypi:~ $ uname -m
aarch64

pi@raspberrypi:~ $ arch
aarch64
```

PromoFaux:
> * `killall` requires the `-s|--signal` flag for any signal execpt `SIGTERM`

Where did you read this?

From here: https://linux.die.net/man/1/killall

> Signals can be specified either by name (e.g. -HUP or -SIGHUP ) or by number (e.g. -1) or by option -s.

PromoFaux:
> Where did you read this?

@yubiuser

yubiuser:
I didn't check the man pages, just `killall --help`. 
I just added this change because at some point dealing with `alpine` based image it was complaining about `killall -9 xxxx` saying something about unknown argument.

But the alpine man page for `killall` also says

```
DESCRIPTION
       killall sends a signal to all processes running any of the specified
       commands.  If no signal name is specified, SIGTERM is sent.

       Signals can be specified either by name (e.g.  -HUP or -SIGHUP) or by
       number (e.g.  -1) or by option -s.

```

rdwebdesign:
Searching the Alpine packages, this is the package for `killall`:
https://pkgs.alpinelinux.org/package/v3.18/main/aarch64/psmisc

This page links to the source code (apparently the same used by Debian).

The source code shows numbers are also accepted as arguments:
https://gitlab.com/psmisc/psmisc/-/blob/master/src/killall.c#L1041-1048

PromoFaux (approved):

PR #1381, opened 2023-07-15
Remove one layer by having both git clone together

Gontier-Julien:
Basically tidy up both git clone together.

## Description

Removed the second 'RUN' git clone, and add it directly after the second one.

## Motivation and Context

It isn't required and doesn't solve any problem, but (maybe this is a old habit of older version) i think it always better to have less layer if possible.
And it doesn't make it less readable so i think it okay

## How Has This Been Tested?

Tested and working, it shouldn't affect the container anyways.

## Types of changes

- [x] Impovement (added my own there but it just a improvement)

## Checklist:

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -31,8 +31,8 @@ RUN apk add --no-cache git \
     less
 
 # download a the main repos from github
-RUN git clone --branch development-v6 https://github.com/pi-hole/AdminLTE.git /var/www/html/admin
-RUN git clone --branch development-v6 https://github.com/pi-hole/pi-hole.git /etc/.pihole
+RUN git clone --branch development-v6 https://github.com/pi-hole/AdminLTE.git /var/www/html/admin &&\
+    git clone --branch development-v6 https://github.com/pi-hole/pi-hole.git /etc/.pihole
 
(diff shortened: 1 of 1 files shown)


src/Dockerfile
```diff
@@ -31,8 +31,8 @@ RUN apk add --no-cache git \
     less
 
 # download a the main repos from github
-RUN git clone --branch development-v6 https://github.com/pi-hole/AdminLTE.git /var/www/html/admin
-RUN git clone --branch development-v6 https://github.com/pi-hole/pi-hole.git /etc/.pihole
+RUN git clone --branch development-v6 https://github.com/pi-hole/AdminLTE.git /var/www/html/admin &&\
```
rdwebdesign, line 34:
```suggestion
RUN git clone --branch development-v6 https://github.com/pi-hole/AdminLTE.git /var/www/html/admin && \
```

Please, add a space between `&&` and `\` for readability (and consistence with other multi-line commands).

rdwebdesign (changes requested):

PromoFaux (approved):
Don't worry about failed builds for now - the script needs tweaking for contributors

PR #1385, opened 2023-07-18
Make the FTL log tail optional. Pass in `TAIL_FTL_LOG: 0` to silence it

PromoFaux:

See title. By default the image will tail the FTL log. Users can disable this if it is too noisy for their preference

---

---

### DIFF
--- a/src/start.sh
+++ b/src/start.sh
@@ -127,7 +127,7 @@ start() {
   capsh --user=$DNSMASQ_USER --keep=1 -- -c "/usr/bin/pihole-FTL $FTL_CMD >/dev/null" &
 
   if [ "${TAIL_FTL_LOG:-1}" -eq 1 ]; then
-    tail -f /var/log/pihole-FTL.log &
+    tail -f /var/log/pihole/FTL.log &
   else
     echo "  [i] FTL log output is disabled. Remove the Environment variable TAIL_FTL_LOG, or set it to 1 to enable FTL log output."
   fi
--- a/src/start.sh
+++ b/src/start.sh
@@ -114,7 +114,7 @@ start() {
   fi
 
   if [ -n "${SKIPGRAVITYONBOOT}" ]; then
-    echo "  Skipping Gravity Database Update."
+    echo "  [i] Skipping Gravity Database Update."
   else
     pihole -g
   fi
@@ -126,7 +126,11 @@ start() {
   sh /opt/pihole/pihole-FTL-prestart.sh
   capsh --user=$DNSMASQ_USER --keep=1 -- -c "/usr/bin/pihole-FTL $FTL_CMD >/dev/null" &
 
-  tail -f /var/log/pihole-FTL.log &
+  if [ "${TAIL_FTL_LOG:-1}" -eq 1 ]; then
+    tail -f /var/log/pihole-FTL.log &
+  else
+    echo "  [i] FTL log output is disabled. Remove the Environment variable TAIL_FTL_LOG, or set it to 1 to enable FTL log output."
+  fi
 
   # https://stackoverflow.com/a/49511035
   wait $!


edgd1er (commented):
If no logs are required may be pihole-FTL could be started et kept in foreground ('-f)

PromoFaux:
If we did that, then the container would stop if ever FTL was restarted/stopped.

For now at least I'd like to keep the command as-is

edgd1er:
Could https://github.com/Yelp/dumb-init be of any help to handle Init 1 pid ? it would send signals (3,9,15) to pihole-FTL requiring a proper process termination and database would stay clean. All process started with the script would be properly terminated also:
(tail, logrotate if any, crond ....)

"Normally, a signal like SIGTERM received by a shell isn't forwarded to subprocesses; instead, only the shell process dies. With dumb-init, you can just write shell scripts with dumb-init in the shebang"

PromoFaux:
Tempted to go back to s6, tbh.

Gontier-Julien:
> Tempted to go back to s6, tbh.
> 
> Edit: That said, something like `dumb-init` or `tini` is probably enough these days.
> 
> `tini` supports more architectures by the looks of things (just going by releases)

One other advantage of `tini` is that it is supported by docker.

From they're readme page:

> NOTE: If you are using Docker 1.13 or greater, Tini is included in Docker itself. This includes all versions of Docker CE. To enable Tini, just [pass the --init flag to docker run](https://docs.docker.com/engine/reference/commandline/run/).

And it even in the alpine repo too.

Maybe you did already know it, but i just wanted to point it out because it neat ^^
i think it would be nice if it is need to use `tini` over `s6` since it also lighter and enough as you said.

PromoFaux:
Yeah, @dschaper already pointed out `tini` to me a while back, but I was hoping (rather naively) that we wouldn't need it, hence getting excited and ripping s6 out when I started on this version of the container!

PromoFaux:
Converted to draft for the time being

PromoFaux:
> If no logs are required may be pihole-FTL could be started et kept in foreground ('-f)

Something else I realised in doing some testing this evening. I am not sure what it is, but I remember it happening on v5 container too, if we start pihole-FTL in the foreground, it stops outputting to the docker log. Note there are more lines in the actual file than there are on the bottom half of the screen:

image

DL6ER:
What is actually puzzling here is that it stops right in the middle of reporting the database statistics. They are simply a collection of `printf` statements, there is nothing in the code that could actually make it stop after the first line:

https://github.com/pi-hole/FTL/blob/5753b50df40defa15d16f6989e12a9ecf3439be9/src/log.c#L502-L512

PromoFaux:
Marked conversation about init systems as outdated due to #1406 being merged. 

Starting FTL in the foreground leads to weirdness with the output where it just cuts out - see previous two comments. This works well enough. Either we tail the log or we don't. Simples.

yubiuser (changes requested):
Needs documentation of the new env setting.

PromoFaux:
@yubiuser  https://github.com/pi-hole/docker-pi-hole/pull/1429

src/start.sh
```diff
@@ -126,7 +126,11 @@ start() {
   sh /opt/pihole/pihole-FTL-prestart.sh
   capsh --user=$DNSMASQ_USER --keep=1 -- -c "/usr/bin/pihole-FTL $FTL_CMD >/dev/null" &
 
-  tail -f /var/log/pihole-FTL.log &
+  if [ "${TAIL_FTL_LOG:-1}" -eq 1 ]; then
+    tail -f /var/log/pihole-FTL.log &
```
yubiuser, line 130: You might want to change to

```suggestion
    tail -f /var/log/pihole/FTL.log &
```

As the other is only a simlink we wanted to remove with v6. See

https://github.com/pi-hole/pi-hole/blob/fef2861eae23442ad823d51c9b4bfdd5b38ab200/advanced/Templates/pihole-FTL-prestart.sh#L29-L34

PromoFaux: Good point well made

yubiuser (approved):

PR #1391, opened 2023-07-19
Reduce git clone size, by cloning only what is needed

Gontier-Julien:
I've managed to reduce the general container size, but telling git not to clone the whole repo.

## Description

When doing a `git clone`, git while clone the whole repo (seem logical haha), but for the container we don't need all of the branch, old commits etc, etc...
So we only tell git clone, to clone what we need (the latest commit and only this branch that we tell it)
edit: For that we use `-depth 1` & `--single-branch` to do it.

## Motivation and Context

It reduce the size of the docker container significantly !
From 117MB (currently from the latest pull) to 73.3MB ! (-43.8MB less)
(i wanted to get the % of it but idk how haha)

## How Has This Been Tested?

Wasn't breaking anything on my side.

## Types of changes

- [x] Improvment

## Checklist:

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -20,8 +20,8 @@ RUN apk add --no-cache \
     procps
 
 # download a the main repos from github
-RUN git clone --branch development-v6 https://github.com/pi-hole/AdminLTE.git /var/www/html/admin && \
-    git clone --branch development-v6 https://github.com/pi-hole/pi-hole.git /etc/.pihole
+RUN git clone --depth 1 --single-branch --branch development-v6 https://github.com/pi-hole/AdminLTE.git /var/www/html/admin && \
+    git clone --depth 1 --single-branch --branch development-v6 https://github.com/pi-hole/pi-hole.git /etc/.pihole
 
 # Download the latest version of pihole-FTL for alpine:
 # Probably need this to be built for different FTLARCHs


PromoFaux:
> From 117MB (currently from the latest pull) to 73.3MB ! (-43.8MB less)

Yeah, I guess downloading the entire history of the web interface repo will do it. The only thing I can think that this will affect is the grabbing the version numbers. What does `pihole -v` output on the console after this change?

(I can test later, just at work currently)

Gontier-Julien:
Still working as excepted
```
  Pi-hole version is development-v6 eda83a4 (Latest: v5.17.1)
  AdminLTE version is development-v6 395ceef (Latest: v5.20.1)
  FTL version is development-v6 vDev-5a6d86a (Latest: v5.23)
```

PromoFaux:
OK, I think that shouldn't be a problem on the `development-v6` branches, however once v6 is released into `master`, then we can add some further logic to make sure that the repo is reset back to the latest tagged commit (that can go hand in hand with: https://github.com/pi-hole/docker-pi-hole/pull/1384

Same thing is done here:

https://github.com/pi-hole/pi-hole/blob/f270642d673d3ff1394bea4f007b65579bca1c84/automated%20install/basic-install.sh#L388

Basically, `master` can sometimes be ahead of the latest release for meta changes such as workflows and the like.

PromoFaux (approved):

PR #1393, opened 2023-07-23
fix arm64 arch.  use copy --chmod. reduce number of layers.

edgd1er:
`{Please select 'base: dev' as target branch above! (you can delete this line)}`

## Description

* replace chmod in run § with COPY --chmod=0755
* reduce number of layers (each RUN is a layer)
* add temporarily binutils to have readelf as sometimes pihole-FTL is not downloaded (html error). I guess this could be removed when the cause wil be identified.
* arm64 arch could not download as the arch name was wrong as the FTLARCH value
* Moved ARG/ENV to top of file. (easier to find where to search when changing values)

## Motivation and Context

* pihole-FTL for arm64 was not downloaded.
* reduce size reducing layers (RUN)

## How Has This Been Tested?

run dig commands on pihole on a amd64 (intel i5) and arm64 (Rpi4)

## Types of changes

## Checklist:

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -1,7 +1,15 @@
 ARG CONTAINER="3.18"
 FROM alpine:${CONTAINER}
 # https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope
+
 ARG TARGETPLATFORM
+ARG WEB_BRANCH="development-v6"
+ARG CORE_BRANCH="development-v6"
+ARG FTL_BRANCH="development-v6"
+
+ENV TZ=Europe/Paris
+ENV DNSMASQ_USER=pihole
+ENV FTL_CMD=no-daemon
 
 RUN apk add --no-cache \
     git \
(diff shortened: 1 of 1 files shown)


src/Dockerfile
```diff
@@ -18,31 +26,28 @@ RUN apk add --no-cache \
...
 
 # download a the main repos from github
-RUN git clone --depth 1 --single-branch --branch ${WEB_BRANCH} https://github.com/pi-hole/AdminLTE.git /var/www/html/admin && \
-    git clone --depth 1 --single-branch --branch ${CORE_BRANCH} https://github.com/pi-hole/pi-hole.git /etc/.pihole
-
-# Download the latest version of pihole-FTL for alpine:
-# Probably need this to be built for different FTLARCHs
-RUN if   [ "$TARGETPLATFORM" = "linux/amd64" ];    then FTLARCH=amd64; \
+RUN git clone --depth 1 --single-branch --branch development-v6 https://github.com/pi-hole/AdminLTE.git /var/www/html/admin && \
+    git clone --depth 1 --single-branch --branch development-v6 https://github.com/pi-hole/pi-hole.git /etc/.pihole ;\
```
rdwebdesign, line 37: Why did you hard-code the branches here?

```suggestion
RUN git clone --depth 1 --single-branch --branch ${WEB_BRANCH} https://github.com/pi-hole/AdminLTE.git /var/www/html/admin && \
    git clone --depth 1 --single-branch --branch ${CORE_BRANCH} https://github.com/pi-hole/pi-hole.git /etc/.pihole ;\
```

rdwebdesign (changes requested):

src/Dockerfile
edgd1er: my mistake, sorry.

src/Dockerfile
```diff
@@ -18,31 +26,28 @@ RUN apk add --no-cache \
...
 # download a the main repos from github
 RUN git clone --depth 1 --single-branch --branch ${WEB_BRANCH} https://github.com/pi-hole/AdminLTE.git /var/www/html/admin && \
-    git clone --depth 1 --single-branch --branch ${CORE_BRANCH} https://github.com/pi-hole/pi-hole.git /etc/.pihole
-
-# Download the latest version of pihole-FTL for alpine:
-# Probably need this to be built for different FTLARCHs
-RUN if   [ "$TARGETPLATFORM" = "linux/amd64" ];    then FTLARCH=amd64; \
+    git clone --depth 1 --single-branch --branch ${CORE_BRANCH} https://github.com/pi-hole/pi-hole.git /etc/.pihole ;\
+    # Download the latest version of pihole-FTL for alpine:
+    # Probably need this to be built for different FTLARCHs
```
PromoFaux, line 39:
```suggestion

```

We can lose this line now

src/Dockerfile
```diff
@@ -1,7 +1,15 @@
 ARG CONTAINER="3.18"
 FROM alpine:${CONTAINER}
 # https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope
+
 ARG TARGETPLATFORM
+ARG WEB_BRANCH="development-v6"
+ARG CORE_BRANCH="development-v6"
+ARG FTL_BRANCH="development-v6"
+
+ENV TZ=Europe/Paris
```
PromoFaux, line 10:
```suggestion
```

This actually does nothing in alpine it seems, however, we can add `tzdata` in a separate pull request, but leave it as default. Users can set the timezone at run time

https://stackoverflow.com/a/68996528

src/Dockerfile
```diff
@@ -53,18 +58,12 @@ RUN cd /etc/.pihole && \
...
-
-ENV DNSMASQ_USER=pihole
-ENV FTL_CMD=no-daemon
-RUN addgroup -S pihole && adduser -S pihole -G pihole
-# RUN groupadd pihole && useradd -r --no-user-group -g pihole -s /usr/sbin/nologin pihole
-
-COPY bash_functions.sh /usr/bin/bash_functions.sh
-COPY start.sh /usr/bin/start.sh
+    install -T -m 0755 ./advanced/Templates/pihole-FTL-poststop.sh /opt/pihole/pihole-FTL-poststop.sh && \
+    # RUN groupadd pihole && useradd -r --no-user-group -g pihole -s /usr/sbin/nologin pihole
```
PromoFaux, line 62:
```suggestion

```

This can go, too. It was a commented out line left over from some earlier experimentation

PromoFaux (changes requested):

PromoFaux (dismissed):

PromoFaux:
Sorry, I should be a stickler here... Can you squash all commits and also sign off per instructions here to keep the bot happy

https://github.com/pi-hole/docker-pi-hole/pull/1393/checks?check_run_id=15273573118

PromoFaux (approved):

PR #1404, opened 2023-07-26
[dev-v6]:extracted uid/gid functions need to be rewritten to remove the return

edgd1er:
`{Please select 'base: dev' as target branch above! (you can delete this line)}`

## Description

do no use return when testing test uid/gid 

## Motivation and Context

modifyUser/Group was a function in pihole v5, now these are in the main part of the script.
Extracted functions need to be rewritten a bit to remove "return"

Not declared as an issue yet.
## How Has This Been Tested?

When PIHOLE_UID and PIHOLE_GID are defined, uid/gid should be redefined:
```
docker-pi-hole-pihole-1  |   [i] Starting docker specific checks & setup for docker pihole/pihole
docker-pi-hole-pihole-1  | + '[' -n 1000 ']'
docker-pi-hole-pihole-1  | ++ id -u
docker-pi-hole-pihole-1  | + currentId=0
docker-pi-hole-pihole-1  |   [i] Changing ID for user: pihole (0 => 1000)
docker-pi-hole-pihole-1  | + [[ 0 -ne 1000 ]]
docker-pi-hole-pihole-1  | + echo '  [i] Changing ID for user: pihole (0 => 1000)'
docker-pi-hole-pihole-1  | + usermod -o -u 1000 pihole
docker-pi-hole-pihole-1  | + '[' -n 1000 ']'
docker-pi-hole-pihole-1  | ++ id -g pihole
docker-pi-hole-pihole-1  | + currentId=101
docker-pi-hole-pihole-1  | + [[ 101 -ne 1000 ]]
docker-pi-hole-pihole-1  | + echo '  [i] Changing ID for group: pihole (101 => 1000)'
docker-pi-hole-pihole-1  |   [i] Changing ID for group: pihole (101 => 1000)
docker-pi-hole-pihole-1  | + groupmod -o -g 1000 pihole
```

## Types of changes

## Checklist:

### DIFF
--- a/src/start.sh
+++ b/src/start.sh
@@ -4,7 +4,6 @@ if [ "${PH_VERBOSE:-0}" -gt 0 ] ; then
     set -x ;
 fi
 
-
 # The below functions are all contained in bash_functions.sh
 # shellcheck source=/dev/null
 . /usr/bin/bash_functions.sh
@@ -27,20 +26,20 @@ echo "  [i] Starting docker specific checks & setup for docker pihole/pihole"
 
 # If PIHOLE_UID is set, modify the pihole user's id to match
 if [ -n "${PIHOLE_UID}" ]; then
-    currentId=$(id -u ${username})
-    [[ ${currentId} -eq ${PIHOLE_UID} ]] && return
-
+  currentId=$(id -u ${username})
+  if [[ ${currentId} -ne ${PIHOLE_UID} ]]; then
     echo "  [i] Changing ID for user: pihole (${currentId} => ${PIHOLE_UID})"
     usermod -o -u ${PIHOLE_UID} pihole
+  fi
 fi
 
 # If PIHOLE_GID is set, modify the pihole group's id to match
 if [ -n "${PIHOLE_GID}" ]; then
-    currentId=$(id -g pihole)
-    [[ ${currentId} -eq ${PIHOLE_GID} ]] && return
-
+  currentId=$(id -g pihole)
+  if [[ ${currentId} -ne ${PIHOLE_GID} ]]; then
     echo "  [i] Changing ID for group: pihole (${currentId} => ${PIHOLE_GID})"
     groupmod -o -g ${PIHOLE_GID} pihole
+  fi
 fi
 
 fix_capabilities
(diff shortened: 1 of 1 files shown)


PromoFaux:
That's fine, we don't need an issue for everything.

Please make the DCO bot happy and then I will merge

src/start.sh
```diff
@@ -27,20 +26,20 @@ echo "  [i] Starting docker specific checks & setup for docker pihole/pihole"
...
 # If PIHOLE_UID is set, modify the pihole user's id to match
 if [ -n "${PIHOLE_UID}" ]; then
-    currentId=$(id -u ${username})
-    [[ ${currentId} -eq ${PIHOLE_UID} ]] && return
-
+  currentId=$(id -u ${username})
+  if [[ ${currentId} -ne ${PIHOLE_UID} ]]; then
     echo "  [i] Changing ID for user: pihole (${currentId} => ${PIHOLE_UID})"
     usermod -o -u ${PIHOLE_UID} pihole
+  fi
```
PromoFaux, line 33: Actually, just for verbosity - can we add  something like

```bash
else
   echo "  [i] ID for user pihole is already ${PIHOLE_UID}, no need to change"
```

Might save some troubleshooting in the future.

Side note: I wonder if we should force the ids to be 999 by default as they are in the v5 containers. I noticed the other day that the id in the new container is 100 - which might cause issues with upgrades?

edgd1er: I've pushed the unchanged info text.
As for the default uid/gid number, start uid depends on distribution, some will start a 100, other at 500, or 1000.
I'm using a ubuntu as host and a debian. I haven't experienced issues with uid/gid other than the "can't chmod <namethefile>". If it needs to be set, Dockerfile would be the good place to implement it when creation the user/group.

PromoFaux (approved):

src/start.sh
```diff
@@ -27,20 +26,24 @@ echo "  [i] Starting docker specific checks & setup for docker pihole/pihole"
 
 # If PIHOLE_UID is set, modify the pihole user's id to match
 if [ -n "${PIHOLE_UID}" ]; then
-    currentId=$(id -u ${username})
```
edgd1er, line 30: a leftover from when it was a function, ${username} should be replaced by pihole

PR #1406, opened 2023-07-26
Ensure pihole-FTL process is shut down cleanly on container exit / add tini

PromoFaux:

While playing around with adding `tini` into the container, I realised that it wasn't a lot of good because `start.sh` will always be the sole child process of  `tini` - and it can only have one child.

As pihole-FTL is spawned as a child of `start.sh`, we would need to add in a trap to `start.sh` anyway, to forward that onto `pihole-FTL`

~~Seems like we can just skip adding `tini` (saving a whopping 2MB in the process 😏) and just trap for exits and shut down `pihole-FTL` first.~~

Revision: Include tini for it's zombie process reaping abilities. If it turns out to be more than needed, we can remove it in a future step

---

---

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -28,6 +28,7 @@ RUN apk add --no-cache \
     psmisc \
     shadow \
     sudo \
+    tini \
     tzdata \
     unzip \
     wget
@@ -71,4 +72,4 @@ COPY --chmod=0755 start.sh /usr/bin/start.sh
 
 HEALTHCHECK CMD dig +short +norecurse +retry=0 @127.0.0.1 pi.hole || exit 1
 
-ENTRYPOINT [ "start.sh" ]
+ENTRYPOINT ["/bin/tini", "--", "start.sh"]
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -72,4 +72,4 @@ COPY --chmod=0755 start.sh /usr/bin/start.sh
 
 HEALTHCHECK CMD dig +short +norecurse +retry=0 @127.0.0.1 pi.hole || exit 1
 
-ENTRYPOINT ["/bin/tini", "--", "start.sh"]
+ENTRYPOINT ["/sbin/tini", "--", "start.sh"]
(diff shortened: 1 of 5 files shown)


edgd1er:
At first, I thought tini was the solution then I read [tini + bash 4.x](https://ahmet.im/blog/minimal-init-process-for-containers/)
Should stop function be the place to handle all processes that were start in the background (logrotate if any, crond, or any other daemon that may come ) ?
what about [zombie child reaping](https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/), could it be something that needs to be handled ? FTL may not have child process, but cron or logrotate may have through the scripts they run ?

PromoFaux:
I have some things to play about with later. The `tini + bash 4.x` link is one of them!

src/Dockerfile
```diff
@@ -27,6 +27,7 @@ RUN apk add --no-cache \
     psmisc \
     shadow \
     sudo \
+    tini \
```
yubiuser, line 30: Didn't you want to save 2 MB?...

yubiuser (commented):
Why are you adding the `uninstall.sh` to the image?

If you want to same even the last bytes you could use a multi-stage build with one stage where you only download the repos to and the final stage where you only the copy the necessary data from the repo....

PromoFaux:
> Why are you adding the `uninstall.sh` to the image?

It's an oversight - up until recently this has been chucking shit at the wall to see what sticks. There is still work to do in optimising the image, but it's being actively developed so, things will change.

The multistage build is something I'm toying with offline - (pretty sure I've mentioned this in one of the issues)

src/Dockerfile
PromoFaux: The smirking emoji was supposed to be representative of my not entire serious nature when saying this.

yubiuser (changes requested):
Merge conflicts ahead

PromoFaux:
Merging sans-approval

PR #1407, opened 2023-07-28
Tell alpine specifically to install/use procps-ng

Gontier-Julien:
## Description

Tell Alpine to specifically install/use `procps-ng` instead of `procps` 

## Motivation and Context

In Alpine `procps` alone doesn't exist there is only `procps-ng` & `procps-compat` (which is the old version of procps).
Telling Alpine just `procps` alone doesn't seem to be an issue to be honest and it will use `procps-ng` automatically, but i guess just for the peace of mind just directly tell it.

</br>
For context about procps-ng:
Procps-ng is a fork of the original procps (which is unmaintained now), and is maintained now by the maintainers
of Debian, Fedora and openSUSE.

https://gitlab.com/procps-ng/procps/-/wikis/home 

## How Has This Been Tested?

It didn't change anything since Alpine already used it alredy.

## Types of changes

## Checklist:

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -25,7 +25,7 @@ RUN apk add --no-cache \
     iproute2-ss \
     jq \
     coreutils \
-    procps \
+    procps-ng \
     ncurses \
     binutils \
     tzdata \


PromoFaux (approved):
Cool 👍  https://pkgs.org/download/procps

dschaper:
> i guess just for the peace of mind just directly tell it.

Yeah, always good to be explicit with dependencies.

PR #1410, opened 2023-07-31
Add PADD to the image

yubiuser:
Add the v6 development version of PADD to the image. 

PADD expects the API at port 80, currently FTL still defaults to 8080. Therefore, PADD needs to be started with `padd --port 8080`

Screenshot at 2023-07-31 23-33-02

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -35,8 +35,7 @@ ADD https://ftl.pi-hole.net/macvendor.db /macvendor.db
 COPY crontab.txt /crontab.txt
 
 # Add PADD to the container, too.
-ADD https://raw.githubusercontent.com/pi-hole/PADD/PADD_FTLv6/padd.sh /usr/local/bin/padd
-RUN chmod +x /usr/local/bin/padd
+ADD --chmod=0755 https://raw.githubusercontent.com/pi-hole/PADD/PADD_FTLv6/padd.sh /usr/local/bin/padd
 
 # download a the main repos from github
 RUN git clone --depth 1 --single-branch --branch ${WEB_BRANCH} https://github.com/pi-hole/AdminLTE.git /var/www/html/admin && \
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -7,6 +7,7 @@ ARG WEB_BRANCH="development-v6"
 ARG CORE_BRANCH="development-v6"
 ARG FTL_BRANCH="development-v6"
 ARG PIHOLE_DOCKER_TAG="unknown"
+ARG PADD_BRANCH="PADD_FTLv6"
 
 ENV DNSMASQ_USER=pihole
 ENV FTL_CMD=no-daemon
@@ -35,7 +36,7 @@ ADD https://ftl.pi-hole.net/macvendor.db /macvendor.db
 COPY crontab.txt /crontab.txt
 
 # Add PADD to the container, too.
-ADD --chmod=0755 https://raw.githubusercontent.com/pi-hole/PADD/PADD_FTLv6/padd.sh /usr/local/bin/padd
+ADD --chmod=0755 https://raw.githubusercontent.com/pi-hole/PADD/${PADD_BRANCH}/padd.sh /usr/local/bin/padd
 
 # download a the main repos from github
 RUN git clone --depth 1 --single-branch --branch ${WEB_BRANCH} https://github.com/pi-hole/AdminLTE.git /var/www/html/admin && \
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -34,6 +34,10 @@ RUN apk add --no-cache \
 ADD https://ftl.pi-hole.net/macvendor.db /macvendor.db
 COPY crontab.txt /crontab.txt
 
+# Add PADD to the container, too.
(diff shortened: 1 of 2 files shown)


src/Dockerfile
```diff
@@ -34,6 +34,10 @@ RUN apk add --no-cache \
 ADD https://ftl.pi-hole.net/macvendor.db /macvendor.db
 COPY crontab.txt /crontab.txt
 
+# Add PADD to the container, too.
+ADD https://raw.githubusercontent.com/pi-hole/PADD/PADD_FTLv6/padd.sh /usr/local/bin/padd
```
edgd1er, line 38: You may spare a layer using:
ADD --chmod=0755 https://raw.githubusercontent.com/pi-hole/PADD/PADD_FTLv6/padd.sh /usr/local/bin/padd

it could be applied an all ADD lines on the dockerfile. Changing owner is also possible:
ADD --chown=user:pihole --chmod=555 https://ftl.pi-hole.net/macvendor.db /macvendor.db

yubiuser: Thanks. I added the alias and included the chmod.

src/Dockerfile
```diff
@@ -34,6 +34,10 @@ RUN apk add --no-cache \
 ADD https://ftl.pi-hole.net/macvendor.db /macvendor.db
 COPY crontab.txt /crontab.txt
 
+# Add PADD to the container, too.
+ADD --chmod=0755 https://raw.githubusercontent.com/pi-hole/PADD/PADD_FTLv6/padd.sh /usr/local/bin/padd
+RUN alias padd='port=${FTLCONF_webserver_port%%,*} padd --port ${port:-8080} --secret ${FTLCONF_webserver_api_password}'
```
PromoFaux, line 39: The alias is not much use here, it will need to be in start.sh like the following (tested)

```bash
# Add an alias for padd to the root user's bashrc
port="${FTLCONF_webserver_port%%,*}"
echo "alias padd='padd --port ${port:-8080} --secret ${FTLCONF_webserver_api_password}'" > /root/.bashrc
```

Another reason for it going into `start.sh` is that the `FTLCONF_` variables will be defined by the user at runtime and we don't know them when we build the image

yubiuser: Thanks, I FP'd it to the right place.

edgd1er: I'm sorry my suggestion was incomplete.
the complete line to put in dockerfile is:
`echo -e "alias padd='port=${FTLCONF_webserver_port%%,*} padd --port ${port:-8080}  --secret ${FTLCONF_webserver_api_password}' >>~/.bashrc"`

without the redirection to bashrc and placed in start.sh will do also.

src/start.sh
```diff
@@ -87,6 +87,10 @@ if [ "${INSTALL_DEV_TOOLS:-0}" -gt 0 ] ; then
     apk add --no-cache nano less
 fi
 
+# Add an alias for padd to the root user's bashrc
```
edgd1er, line 90: placed in start.sh will do also, but bashrc will be rewritten at each container start, erasing possible user aliases and/or bash user defined settings.

PromoFaux: It needs to be in `start.sh` because the user-defined `FTLCONF_` environment variables will not be defined at build-time, so it need to be applied at run time.

@yubiuser another thought, to ensure we're not munging any user-defined `.bashrc` (though, one would argue there isn't much of a need to do so in a container such as this) Perhaps we cna do something like:

Dockerfile:
```dockerfile
ADD --chmod=0755 https://raw.githubusercontent.com/pi-hole/PADD/PADD_FTLv6/padd.sh /usr/bin/padd
```

start.sh
```bash
port="${FTLCONF_webserver_port%%,*}"
echo "/usr/bin/padd --port ${port:-8080} --secret ${FTLCONF_webserver_api_password}'" > /usr/local/bin/padd
chmod +x /usr/local/bin/padd
```

PromoFaux: > using $ prevent bash to substitute variable at buildtime.

Ah OK thanks, when I was tinkering about before it wasn't working - but that could have been for all manner of reasons, to be honest!

edgd1er: I've just noticed that the slash before the $ was removed in the reply when publishing the comment, there is may be a misunderstanding : here is a picture of the dockerfile of how it should look:
image

and the result in the image:
image

PromoFaux:
With the addition of https://github.com/pi-hole/PADD/pull/362,  the user need only pass in `--secret` to padd now in order for it to work.

Internally discussed, but as there is no guarantee that the password will be set via `FTLCONF_webserver_api_password`, (i.e some users prefer not to keep passwords in their compose files etc) it does not seem like a good idea to have an alias for padd inside the image. We can always add a hint in the readme on how users can add an alias themselves if they really need it

@yubiuser -can we have it just in the dockerfile please, no alias - and also if possible please make the padd branch a variable that can be passwed in at build time. (like the other repo branch arguments)

src/start.sh
```diff
@@ -87,6 +87,7 @@ if [ "${INSTALL_DEV_TOOLS:-0}" -gt 0 ] ; then
     apk add --no-cache nano less
 fi
 
+
```
PromoFaux, line 90:
```suggestion

```

PromoFaux (approved):

PR #1412, opened 2023-08-02
Make docker-pihole posix compliant.

Gontier-Julien:
⚠️**This is a very much work in progress do NOT merge until everything as been tested**⚠️
## Description

This is the start on trying (at least) to get first everything in `docker-pihole` posix compliant, then in the main repo (`pi-hole/pi-hole`).

I'm very bound to make some mistake in this transition from bash to posix.

Any help is very much appreciated ! ^^

## Motivation and Context

This work is just to try to make pi-hole posix compliant, this could/would make pi-hole _easier_ to port to different system and _should_ make it behave as excepted in all the supported OS.

_Also i think this could have it very own branch rather than development-v6_

## How Has This Been Tested?

At the moment **ABSOLUTELY NOT** it very much a work in progress.

## Types of changes

## Checklist:

### DIFF
--- a/build-and-test.sh
+++ b/build-and-test.sh
@@ -1,7 +1,7 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 set -ex
 
-if [[ "$1" == "enter" ]]; then
+if [ "$1" = "enter" ]; then
     enter="-it --entrypoint=sh"
 fi
 
@@ -11,7 +11,7 @@ GIT_TAG="${GIT_TAG:-$GIT_BRANCH}"
 PLATFORM="${PLATFORM:-linux/amd64}"
 
 # generate and build dockerfile
-docker buildx build --load --platform=${PLATFORM} --tag image_pipenv --file test/Dockerfile test/
+docker buildx build --load --platform="${PLATFORM}" --tag image_pipenv --file test/Dockerfile test/
 docker run --rm \
     --volume /var/run/docker.sock:/var/run/docker.sock \
     --volume "$(pwd):/$(pwd)" \
@@ -19,4 +19,4 @@ docker run --rm \
     --env PIPENV_CACHE_DIR="$(pwd)/.pipenv" \
     --env GIT_TAG="${GIT_TAG}" \
     --env PY_COLORS=1 \
-    ${enter} image_pipenv
+    "${enter}" image_pipenv
--- a/build-and-test.sh
+++ b/build-and-test.sh
@@ -19,4 +19,4 @@ docker run --rm \
     --env PIPENV_CACHE_DIR="$(pwd)/.pipenv" \
     --env GIT_TAG="${GIT_TAG}" \
     --env PY_COLORS=1 \
-    "${enter}" image_pipenv
+    ${enter} image_pipenv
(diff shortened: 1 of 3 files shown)


Gontier-Julien:
PS: i'm using `shellcheck` to try to make it posix compliant.

PromoFaux:
Hmm, CI struggling with `build-and-test.sh`. Is it running OK locally?

Gontier-Julien:
> Hmm, CI struggling with `build-and-test.sh`. Is it running OK locally?

I did not test anything yet, but i will now.

build-and-test.sh
```diff
@@ -11,12 +11,12 @@ GIT_TAG="${GIT_TAG:-$GIT_BRANCH}"
...
+docker buildx build --load --platform="${PLATFORM}" --tag image_pipenv --file test/Dockerfile test/
 docker run --rm \
     --volume /var/run/docker.sock:/var/run/docker.sock \
     --volume "$(pwd):/$(pwd)" \
     --workdir "$(pwd)" \
     --env PIPENV_CACHE_DIR="$(pwd)/.pipenv" \
     --env GIT_TAG="${GIT_TAG}" \
     --env PY_COLORS=1 \
-    ${enter} image_pipenv
+    "${enter}" image_pipenv
```
PromoFaux, line 22: It's this line.

Output from running `build-and-test.sh` on this branch:

```
PY_COLORS=1  image_pipenv
```

Output from another branch

```
PY_COLORS=1 image_pipenv
```

Note the additional space that has been introduced

Gontier-Julien:
Should be okay now, i think

Gontier-Julien:
To make `start.sh` fully posix compliant, all the `RANDOM` need to be replaced.

One solution could be this:
`awk -v min=1 -v max=30000 'BEGIN{srand(); print int(min+rand()*(max-min+1))}'`

The only problem with this one, is that if the command is run with the same second, it while output exactly the same.

Gontier-Julien:
With this, only a few thing left in `bash_functions.sh`, could be renamed `sh_functions.sh afterwards haha

But if someone could see the 3 things left in it, i would appreciate it!

With this i'm going to sleep, maybe some sleep while help me on it haha

PromoFaux:
Sometimes I feel it more effort than it is worth. Especially within the docker container anyway. `bash` is less than a megabyte at last check - and most of the scripts within this container are unique to the container itself, so portability isn't really an issue. Any system that has `docker` or `podman` or something on it should be able to "just" run the container.

>  only a few thing left in bash_functions.sh

Except it is all horribly broken and the image doesn't start. But sure, the shell code itself is posix compliant  😉 

But thanks for giving it a go!

Gontier-Julien:
> Sometimes I feel it more effort than it is worth. Especially within the docker container anyway. `bash` is less than a megabyte at last check - and most of the scripts within this container are unique to the container itself, so portability isn't really an issue. Any system that has `docker` or `podman` or something on it should be able to "just" run the container.
> 
> > only a few thing left in bash_functions.sh
> 
> Except it is all horribly broken and the image doesn't start. But sure, the shell code itself is posix compliant wink
> 
> But thanks for giving it a go!

I mean you right haha, and there still a lot of work on the script side of pi-hole it self.
I'll see if i continue it anyways or not we'll see if i have the motivation x)
At least it a start to somewhere haha

Gontier-Julien:
I feel like it more effort than it is worth too so closing this.

PR #1415, opened 2023-08-09
 Always update to get the latest list of available packages. 

Gontier-Julien:
Always update Alpine APK repo list to get the latest list of available packages. 

## Description

 Always update Alpine APK to have the latest list of available packages to be safe.

## Motivation and Context

Just a very small improvement, let Alpine APK update the list of available packages before installing any.
Just to be on the safe side as always so Alpine always as a fresh list.

## How Has This Been Tested?

Tested and working.

## Types of changes

- [X] Improvements

## Checklist:

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -12,7 +12,7 @@ ARG PADD_BRANCH="PADD_FTLv6"
 ENV DNSMASQ_USER=pihole
 ENV FTL_CMD=no-daemon
 
-RUN apk add --no-cache \
+RUN apk add --update --no-cache \
     bash \
     bind-tools \
     binutils \


dschaper:
`apk add --no-cache` invalidates the cache and, by design, that runs an update. No need for this change.


https://stackoverflow.com/a/49119046
https://docs.alpinelinux.org/user-handbook/0.1a/Working/apk.html

dschaper:
You can see the first step of the RUN command output shows the indexes are fetched.

https://github.com/pi-hole/docker-pi-hole/actions/runs/5792940168/job/15699977214#step:7:175

```bash
#9 [2/9] RUN apk add --no-cache     bash     bind-tools     binutils     coreutils     curl     git     iproute2-ss     jq     libcap     ncurses     nmap-ncat     procps-ng     psmisc     shadow     sudo     tzdata     unzip     wget
#9 0.068 fetch https://dl-cdn.alpinelinux.org/alpine/v3.18/main/x86_64/APKINDEX.tar.gz
#9 0.403 fetch https://dl-cdn.alpinelinux.org/alpine/v3.18/community/x86_64/APKINDEX.tar.gz
#9 0.655 (1/56) Installing ncurses-terminfo-base (6.4_p20230506-r0)
#9 0.661 (2/56) Installing libncursesw (6.4_p20230506-r0)
#9 0.669 (3/56) Installing readline (8.2.1-r1)
#9 0.676 (4/56) Installing bash (5.2.15-r5)
#9 0.691 Executing bash-5.2.15-r5.post-install
#9 0.693 (5/56) Installing fstrm (0.6.1-r2)
#9 0.697 (6/56) Installing krb5-conf (1.0-r2)
#9 0.700 (7/56) Installing libcom_err (1.47.0-r2)
#9 0.703 (8/56) Installing keyutils-libs (1.6.3-r3)
#9 0.707 (9/56) Installing libverto (0.3.2-r2)
#9 0.711 (10/56) Installing krb5-libs (1.20.1-r1)
#9 0.732 (11/56) Installing nghttp2-libs (1.55.1-r0)
#9 0.738 (12/56) Installing protobuf-c (1.4.1-r2)
#9 0.742 (13/56) Installing libuv (1.44.2-r2)
#9 0.748 (14/56) Installing xz-libs (5.4.3-r0)
#9 0.754 (15/56) Installing libxml2 (2.11.4-r0)
```

Gontier-Julien:
Oh didn't know that actually! But good to know then!
Closing this as it not needed.

edgd1er:
@dschaper 
though an image upgrade may be interesting: `apk upgrade --no-cache`
https://stackoverflow.com/questions/72045168/is-it-necessary-to-run-apk-update-apk-upgrade-in-a-docker-build-stage/76440791#76440791

I'm just descoring that aspect and need to read more about it:
`apk upgrade --no-cache && apk add --no-cache --update `
might not be the most efficient...

dschaper:
What is the end goal of running upgrade? If you want to have a solid, reproducible build then doing the opposite is apropos. Pin all the packages to known working versions.

Having a build process that doesn't really know what version of the OS nor the packages will lead to future headache.

Gontier-Julien:
> What is the end goal of running upgrade? If you want to have a solid, reproducible build then doing the opposite is apropos. Pin all the packages to known working versions.
> 
> Having a build process that doesn't really know what version of the OS nor the packages will lead to future headache.
> 
> Edit: I'm not sure of the utility of running upgrade no cache since the next step of add no cache already invalidates the cache so nothing from upgrade would survive?

I agree with @dschaper doing an upgrade won't help reproduce an issue, especially since we already add packages that already defeat it a little.
And Alpine Linux already have a point release every 2 month-ish it seems (`3.18.1`, `3.18.2`, `3.18.3`, etc..) so to be honest it not necessary to do an upgrade. And to add to it, the Alpine docker image it self is regularly build again with the latest version.

edgd1er:
> What is the end goal of running upgrade? If you want to have a solid, reproducible build then doing the opposite is apropos. Pin all the packages to known working versions.

Upgrade will install latest version for already installed packages, in the same alpine release. 3.18 will stay 3.18, as to change release version other actions are [required](https://wiki.alpinelinux.org/wiki/Include:Upgrading_to_latest_release)
As pointed out, in a previous comment, alpine is releasing every two month an update image. worst case scenario, pihole could be vulnerable between two alpine updates, provided pihole image is rebuilt just after the alpine release.
If pihole releases are not synchronized with alpine releases, applying latest fixes would let the built image be as secured as possible at build time.

Docker has a tool to scan published image for basic security scanning, never used it though: https://docs.docker.com/docker-hub/vulnerability-scanning/

as for the `apk add --no-cache --update`  --update is not needed as `--no-cache` implies a `--update`. --no-cache is enough, as it is now.

dschaper:
> Upgrade will install latest version for already installed packages, in the same alpine release.

Ah, okay. This is already covered with the use of `FROM: alpine 3.18` https://github.com/pi-hole/docker-pi-hole/blob/development-v6/src/Dockerfile#L2

That image is pulled on each build, and the images are rebuilt frequently. Because we use `3.18` as the tag we get all of the point releases automatically as they are made. Our docker images are also built on schedule so `nightly` and similar are always the latest. Snow mentions this as well.

I don't think adding in an additional `upgrade` of existing packages is helpful. I concede that it doesn't add much in the way of space or time to do but I think it just duplicates what is already effectively done.

PR #1457, opened 2023-10-16
!!!BREAKING CHANGE!!! use `printenv` instead of `declare` when converting `FTLCONF_` environment variables

PromoFaux:

As an alternative to https://github.com/pi-hole/docker-pi-hole/pull/1456

In this case, rather than the user having to manually replace `.`'s with `_`s in the env var name (e.g `dns.upstreams` becomes `FTLCONF_dns_upstreams`), with this proposed change, they can now pass it in with the `.`s (e.g `dns.upsteams` becomes `FTLCONF_dns.upstreams`)

The advantage of doing this is that we don't have to add in a special case to handle FTL config names that have `_`s built into them as and when they are added.

The disadvantage of this is that merging this change _will_ break any existing `development-v6` containers that are running in userland.

Discuss

---

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ services:
     restart: unless-stopped
 ```
 2. Run `docker compose up -d` to build and start pi-hole (Syntax may be `docker-compose` on older systems)
-3. If using Docker's default `bridge` network setting, set the environment variable `FTLCONF_dns_listeningMode` to `all`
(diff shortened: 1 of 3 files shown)


PromoFaux:
Closing in favour of https://github.com/pi-hole/docker-pi-hole/pull/1456.

We can revisit this at another time if needs be, but it does feel counterintuitive to allow `.`s in an environment variable key name

PR #1460, opened 2023-10-20
Updates to reflect FTL's native capability of reading the FTLCONF_ env vars

PromoFaux:

Now that https://github.com/pi-hole/FTL/pull/1679#pullrequestreview-1689702823 has been merged, we no longer need to parse the environment variables in the startup script (with the exception of `FTLCONF_webserver_api_password`, which we will set to a random value if none is specified) and totally negates the need for the changes made in https://github.com/pi-hole/docker-pi-hole/pull/1456 to add special cases for FTL config names with `_` in them

#### Env var set (With or without the addition of `FTLCONF_ENV_ONLY`)
```
ph-development-v6  |   [i] Checking web password
ph-development-v6  |   [i] Assigning password defined by Environment Variable
```

#### No env var set and `pwhash` empty in config file
```
ph-development-v6  |   [i] Checking web password
ph-development-v6  |   [i] No password set in environment or config file, assigning random password: iK1me016
ph-development-v6  |   [✓] New password set
```

#### No env var set and `pwhash` has a value in config file
```
ph-development-v6  |   [i] Checking web password
ph-development-v6  |   [i] Password already set in config file
```

#### No env var set, but `FTLCONF_ENV_ONLY` set to `true`
```
ph-development-v6  |   [i] Checking web password
ph-development-v6  |   [i] No password supplied via FTLCONF_webserver_api_password, but FTLCONF_ENV_ONLY is set to true, using default (none)
```

We also check the value of `FTLCONF_dns_upstreams` for `;` and replace them with `,` to ensure compatibility with FTL's environment variable parsing. A notice will be displayed to the user in this case.

```
ph-development-v6  |   [i] Replacing ';' with ',' in FTLCONF_dns_upstreams - please update the environment variable to use commas instead of semicolons
```

Lastly some formatting/old code removal has also been applied

---

---


PromoFaux:
Forgot the tests. PR now ready

src/start.sh
```diff
@@ -49,34 +46,16 @@ start() {
...
-  # setup_blocklists
-
-  # FTL setup
-  # ===========================
-
-  # setup_FTL_User
-  # setup_FTL_query_logging
-
-  [ -f /.piholeFirstBoot ] && rm /.piholeFirstBoot
+  # If FTLCONF_dns_upstreams contains any semicolons, replace them with commas
```
yubiuser, line 49: Do we really need this? We could fail hard here, make the user change from semicolons to commas and remove the hand-holding in a few versions.

yubiuser (commented):
L70 of the README: add port 443

src/start.sh
PromoFaux: I guess no, not strictly - but I figured it was a more friendly way of just breaking the containers that are already out there.

DL6ER: FTL can accept `;` as well, we just make

https://github.com/pi-hole/FTL/blob/4eca25e2a5700a35523fc62560447082964d00ae/src/config/toml_helper.c#L936

``` c
const char delim[] =",;";
```

README.md
```diff
@@ -19,39 +19,13 @@
...
-    # Volumes store your data between container upgrades
-    volumes:
-      - './etc-pihole:/etc/pihole'
-     # - './etc-dnsmasq.d:/etc/dnsmasq.d' # Only needed if you have some custom configs for dnsmasq
-    # https://github.com/pi-hole/docker-pi-hole#note-on-capabilities
-    cap_add:
-      - NET_ADMIN # Required if you are using Pi-hole as your DHCP server, else not needed
-    restart: unless-stopped
-```
+1. Copy [the example docker compose file](/examples/docker-compose-example.yml) and update as needed
```
DL6ER, line 24: I typically find it nicer if READMEs have the `docker compose` example right there because you can find it more easily when skimming over it.

yubiuser:
It's not the `arm64` tests that misreport, but `amd64`. 

The reported issue is

```
chown: cannot access '/etc/pihole/logrotate': No such file or directory\r\nchmod: cannot access '/etc/pihole/logrotate': No such file or directory
```
This comes from 

https://github.com/pi-hole/docker-pi-hole/blob/314ed3cd19f3f364441cdd0783c88308bf466014/src/start.sh#L120

Which triggers

https://github.com/pi-hole/pi-hole/blob/6cf39d9b924adf2c8bd49c7f4997b2ffff9e0a53/advanced/Templates/pihole-FTL-prestart.sh#L19-L21

Those lines were just added recently by https://github.com/pi-hole/pi-hole/pull/5444.

So the test are failing rightfully, because on docker, we don't copy `/etc/pihole/logrotate`.  You might want to touch these files on docker and add `arm64` back ;-)

___

(No idea why `adm64` passed, but they should not.

yubiuser:
My previous assumption was wrong, but lead me to find out what really went wrong. It's the check for a set password using `FTL` itself.

https://github.com/pi-hole/docker-pi-hole/blob/98049d19c78078a602f9f2e224c70734e762c3cd/src/bash_functions.sh#L75

It only checks, if there is an response at all. This is the case if FTL returns the password, but also **on errors**. The underlying issue here is: the test pulled the wrong architecture, because `TARGETPLATFORM` in the Dockerfile was not set.

I forked the repo and fixed the tests:

https://github.com/yubiuser/docker-pi-hole/pull/1/commits/7e8712db4025d79cc89eb5f26b1d63389dd61e40

https://github.com/yubiuser/docker-pi-hole/actions/runs/6597244291

___

I think you should re-add the tests, they helped to identify this bug.
Additionally, you could not only check for non-emptyness, but for a string starting with

```
nanopi@nanopi:~$ pihole-FTL --config webserver.api.pwhash
$BALLOON-SHA256....
```

PromoFaux:
Thanks for looking at that @yubiuser - reverted the commit removing the arm tests and added in the additional commit from your fork

README.md
PromoFaux: Restored

src/start.sh
PromoFaux: However you prefer

DL6ER: Less code over there, fewer migration things to handle correctly by just supporting this: https://github.com/pi-hole/FTL/pull/1690

PromoFaux: I will try this once it builds and force push away the change from this side

src/bash_functions.sh
```diff
@@ -71,6 +60,41 @@ ensure_basic_configuration() {
...
+    if [ -z "${FTLCONF_webserver_api_password+x}" ]; then
+        # is the variable FTLCONF_ENV_ONLY set to true?
+        if [ "${FTLCONF_ENV_ONLY}" == "true" ]; then
+            echo "  [i] No password supplied via FTLCONF_webserver_api_password, but FTLCONF_ENV_ONLY is set to true, using default (none)"
+            # If so, return - the password will be set to FTL's default (no password)
+            return
+        fi
+
+        # Exit if password is already set in config file
+        if [[ -n $(pihole-FTL --config webserver.api.pwhash) ]]; then
```
yubiuser, line 75: Still think this should be something like

```suggestion
        if [[  $(pihole-FTL --config webserver.api.pwhash) = \$BALLOON-SHA256* ]]; then
```

README.md
PromoFaux: 👍

src/bash_functions.sh
PromoFaux: Is there an advantage to checking for a value that begins `$BALLOON-SHA256`, vs checking for an empty value?

yubiuser: It's more precis - we want to know the command returns a password hash not anything else (e.g. some kind of API error)

PromoFaux: Cool, fixed in latest commit

DL6ER (approved):
Seems to work as expected

PR #1467, opened 2023-10-29
Allow setting an additional command to run on container start

PromoFaux:

I'm not 100% sure on this one, but I'm not going to throw the idea away immediately without first some discussion around it.

My main use case is per the comments added in `start.sh`, namely:

```bash
# Example use case: If mounting the web repo from the host, we need to run `git config --global --add safe.directory /var/www/html/admin`
#                   in order for the updatechecker to work
```

By setting `ADDITIONAL_COMMAND: "git config --global --add safe.directory /var/www/html/admin"`, I can then bypass the following message when mounting into `/var/www/html/admin`. This is not a common occurrence in userland, so seems silly to run that command regardless (though, potentially OK?)

image

Also allows for silly things like

image

I don't think this is a variable that needs documenting. Or maybe it does. Discuss.

---

---

### DIFF
--- a/src/start.sh
+++ b/src/start.sh
@@ -4,6 +4,14 @@ if [ "${PH_VERBOSE:-0}" -gt 0 ]; then
   set -x
 fi
 
+# Undocumented environment variable to allow easy injection of arbitrary commands.
+# HERE BE DRAGONS!
+# Example use case: If mounting the web repo from the host, we need to run `git config --global --add safe.directory /var/www/html/admin`
+#                   in order for the updatechecker to work
+if [ -n "${ADDITIONAL_COMMAND}" ]; then
+  eval "${ADDITIONAL_COMMAND}"
+fi
+
 trap stop TERM INT QUIT HUP ERR
 
 start() {


yubiuser:
I'm not a big fan of allowing arbitrary command injections during container startup on a deployed user-facing image. 

I'm also not sure why you experienced the `safe.directory` issue. We got some issue reports on `core` about this a year ago since `git` started to check for the owner of the respective git repo. However, this should not be an issue here, as we run `pihole updatechecker` as `root` within the container and `root` should have enought priviledges to access `.git`

PromoFaux:
Yeah, it feels icky to be honest, was just throwing it against the wall.

If you mount an external git repo of the web repo into your container, do you see the same `safe.directory` issue?

yubiuser:
> If you mount an external git repo of the web repo into your container, do you see the same safe.directory issue?

Yes.


```
[✓] Cleaning up stray matter
pihole  | 
pihole  |   [✓] Done.
pihole  | fatal: detected dubious ownership in repository at '/var/www/html/admin'
pihole  | To add an exception for this directory, call:
pihole  | 
pihole  | 	git config --global --add safe.directory /var/www/html/admin
pihole  | fatal: detected dubious ownership in repository at '/var/www/html/admin'
pihole  | To add an exception for this directory, call:
pihole  | 
pihole  | 	git config --global --add safe.directory /var/www/html/admin
pihole  |   [i] Docker start setup complete
pihole  | 
pihole  |   [i] pihole-FTL (no-daemon) will be started as pihole


```

PR #1468, opened 2023-10-30
drop the `/v8` from `linux/arm64/v8`

PromoFaux:

CI is using `linux/arm64/v8` in the `platform` matrix, we should be using `linux/arm64`. I think. Lets see what the tests say...

---

---

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -17,7 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         registry: [dockerhub, ghcr]
-        platform: [linux/amd64, linux/386, linux/arm/v6, linux/arm/v7, linux/arm64/v8]
+        platform: [linux/amd64, linux/386, linux/arm/v6, linux/arm/v7, linux/arm64]
         container: [3.18]
         include:
           - registry: dockerhub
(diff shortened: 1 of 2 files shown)


PromoFaux:
Looks good so far: 

image

PromoFaux:
Actually might have one more commit for this.. hold reviews please

PromoFaux:
Ready for review

yubiuser:
I wonder if we now still need both `TARGETPLATFORM` set from https://github.com/pi-hole/docker-pi-hole/commit/5c7ff5c539c75c1aab7fcd3d392f78d26e246d75

(I get always confused by the nested docker build on top of docker builds)

PromoFaux:
That was what my deleted comment was about.. I tried to build locally with that removed (run as `PLATFORM=linux/arm64/v8 ./build-and-test.sh` and the result was the following:

image

So that part is still needed I believe.

yubiuser (approved):

PromoFaux:
Both `--platform=linux/arm64` and `--platform=linux/arm64/v8` are valid and will produce a `linux/arm64` image (per [the link you shared earlier](https://github.com/BretFisher/multi-platform-docker-build#recommended-approach-for-curl-and-wget-commands-in-multi-platform-dockerfiles)

The docker-in-docker stuff probably adds some complexity, and something we should look at finding another way around one day. But not this day. For now we can just use the simple form of the `platform` argument

PromoFaux:
Watching the next publish like a hawk...

PR #1469, opened 2023-10-30
Fail hard if TARGETPLATFORM does not meet one of the predefined values

PromoFaux:

We probably _shouldn't_ fall back to amd64 if the TARGETPLATFORM is unknown.

Convince me otherwise

Also shuffles the order of the first RUN command around a little. First we attempt to download FTL, and only then we clone the repos.

running something like `docker buildx build ./src/ --platform=linux/s390x ` will cause the `exit 1`

---

---

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -40,22 +40,23 @@ COPY crontab.txt /crontab.txt
 # Add PADD to the container, too.
 ADD --chmod=0755 https://raw.githubusercontent.com/pi-hole/PADD/${PADD_BRANCH}/padd.sh /usr/local/bin/padd
 
-# download a the main repos from github
-RUN git clone --depth 1 --single-branch --branch ${WEB_BRANCH} https://github.com/pi-hole/web.git /var/www/html/admin && \
-    git clone --depth 1 --single-branch --branch ${CORE_BRANCH} https://github.com/pi-hole/pi-hole.git /etc/.pihole ;\
-    # Download the latest version of pihole-FTL for alpine:
-    if   [ "$TARGETPLATFORM" = "linux/amd64" ];    then FTLARCH=amd64; \
+
+# Download the latest version of pihole-FTL for alpine:
+RUN if   [ "$TARGETPLATFORM" = "linux/amd64" ];    then FTLARCH=amd64; \
     elif [ "$TARGETPLATFORM" = "linux/386" ];      then FTLARCH=386; \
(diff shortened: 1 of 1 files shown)


src/Dockerfile
```diff
@@ -39,22 +39,23 @@ COPY crontab.txt /crontab.txt
...
-    git clone --depth 1 --single-branch --branch ${CORE_BRANCH} https://github.com/pi-hole/pi-hole.git /etc/.pihole ;\
-    # Download the latest version of pihole-FTL for alpine:
-    if   [ "$TARGETPLATFORM" = "linux/amd64" ];    then FTLARCH=amd64; \
+
+# Download the latest version of pihole-FTL for alpine:
+RUN if   [ "$TARGETPLATFORM" = "linux/amd64" ];    then FTLARCH=amd64; \
     elif [ "$TARGETPLATFORM" = "linux/386" ];      then FTLARCH=386; \
     elif [ "$TARGETPLATFORM" = "linux/arm/v6" ];   then FTLARCH=armv6; \
     elif [ "$TARGETPLATFORM" = "linux/arm/v7" ];   then FTLARCH=armv7; \
     # Note for the future, "linux/arm6/v8" is not a valid value for TARGETPLATFORM, despite the CI platform name being that.
```
yubiuser, line 48: Maybe move and change that comment to a link with all the valid TARGETPLATFORM values?

src/Dockerfile
```diff
@@ -39,22 +39,23 @@ COPY crontab.txt /crontab.txt
...
     elif [ "$TARGETPLATFORM" = "linux/arm64" ];    then FTLARCH=arm64; \
     elif [ "$TARGETPLATFORM" = "linux/riscv64" ];  then FTLARCH=riscv64; \
-    else FTLARCH=amd64; fi \
+    else echo "FTL Not available for this platform" && exit 1; fi \
     && echo "Arch: ${TARGETPLATFORM}, FTLARCH: ${FTLARCH}" \
```
yubiuser, line 52: I know you want a small image and only a few `RUN` commands. But with stuffing all of this together, we loose readability a lot. I would separate the FTL downloading from repo downloading like we had it before.

src/Dockerfile
PromoFaux: I'm just going to remove the comment.

yubiuser: Do you want to do anything with this comment? Remove? Change?

PR #1470, opened 2023-10-30
Install logrotate

yubiuser:
When starting with an empty `/etc/pihole`, docker startup errors with 

```
pihole  |   [i] Applying the following caps to pihole-FTL:
pihole  |         * CAP_CHOWN
pihole  |         * CAP_NET_BIND_SERVICE
pihole  |         * CAP_NET_RAW
pihole  | 
pihole  | chown: cannot access '/etc/pihole/logrotate': No such file or directory
pihole  | chmod: cannot access '/etc/pihole/logrotate': No such file or directory

```

Reason is, that `/etc/pihole/logrotate` does not exit, but `pihole-FTL.prestart.sh` tries to change owner/permissions.
This PR copies the file from the repo into the image.

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -22,6 +22,7 @@ RUN apk add --no-cache \
     iproute2-ss \
     jq \
     libcap \
+    logrotate \
     ncurses \
     nmap-ncat \
     procps-ng \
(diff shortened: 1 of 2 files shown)


yubiuser:
P.S. The logrotate currently does nothing. This PR is only to prevent the error and can be used as a first step for logrotate. But is logrotate on Docke really necessary?

PromoFaux:
This has been on my backlog for a while. This PR adds the files, but think we may also need to add in the `logrotate` binary as we have with `cron`. I never got much further than starting to think about it....

yubiuser:
I've added `logrotate` to the installed packages. In theory, logrotate `FTL.log` and `pihole.log` should work now already. We have `´cron` installed, it's called on `contab.txt` which executes `pihole flush once quiet` which in turn invokes `logrotate` on `/etc/pihole/logrotate`

PromoFaux (approved):
Lets try it and see what happens

PR #1497, opened 2023-12-08
Update alpine to 3.19

yubiuser:
Updates the base alpine image to `3.19`.

There is no v3.19 for `docker-cli` yet. I kept the v3.18 base there. This PR can be extended later when such an image comes available - or it could be handled in an other PR


yubiuser:
alpine 3.19 is now also available for `docker-cli`. I updated the workflow. PR is ready for review.

yubiuser:
Latest commit removed `pipenv` and allowed `pip` to "break" the system python packages. Python is enforcing the use of venv but setting this up increases the code complexity within our docker-on-docker test suite and makes it even harder to understand. As the container are disposable and only used to run the test, I see no issue with allowing `pip` to ` --break-system-packages`.

DL6ER (approved):

PR #1500, opened 2023-12-10
Enhance theme validation

ADawidowski:
Makes the web theme name validation dynamic based on the existing theme css file names and applies a regexp sanity check just to be sure.

## Description
Modified bash_funcitons.sh:setup_web_theme.sh to replace static web theme name validation with dynamic one in order to make the validation future proof.

## Motivation and Context
Adding new themes requires modification of the validation function. This change modifies the validation code so that it accepts every theme that exists in the file system location.
Example: currently the recently-added lcars-picard theme is not accepted by the validation function because it has not been explicitly added to the validation code. This code change fixes that.

## How Has This Been Tested?
Ran standalone unit tests on the extracted setup_web_theme function. Added tests to the test suite. Added test fixture that temporarily creates empty theme files during the tests for the purpose of testing new known names.

## Types of changes

## Checklist:
- [?] My code follows the code style of this project.

### DIFF
--- a/src/s6/debian-root/usr/local/bin/bash_functions.sh
+++ b/src/s6/debian-root/usr/local/bin/bash_functions.sh
@@ -416,19 +416,16 @@ setup_web_port() {
 }
 
 setup_web_theme(){
-    # Parse the WEBTHEME variable, if it exists, and set the selected theme if it is one of the supported values.
-    # If an invalid theme name was supplied, setup WEBTHEME to use the default-light theme.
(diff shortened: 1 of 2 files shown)


PromoFaux (approved):
Thanks for the submission, and thanks for adding comprehensive tests!

I'll accept this as-is for the time being, but there are no _major_ plans to do another release of the v5-based code any time soon.

In the v6 (see branch `development-v6` all repos) container, all settings are handled internally by `pihole-FTL`, and as such are to be configured with an env var like `FTLCONF_webserver_interface_theme: 'default-dark'` etc.

PR #1511, opened 2023-12-22
Fix v6 build on riscv64

yubiuser:
https://github.com/pi-hole/docker-pi-hole/pull/1497 changed the base alpine version of the docker image to `3.19`. 
In this PR I also renamed the build-arg from `container` to `alpine_version`. However, I did not adjust the workflow file - so the `riscv64` [builds failed ](https://github.com/pi-hole/docker-pi-hole/actions/runs/7300747174/job/19896020464) because there is no `3.19` but only `edge`

https://hub.docker.com/r/riscv64/alpine/


PromoFaux (approved):
Ah yes, GHA should also have been changed to reflect the new argument name.

PR #1526, opened 2024-01-06
Consider registry when up/downloading artifacts

yubiuser:
Include the used registry in the artifact upload name and separate downloads for registries. This hopefully fixes the issue we saw here: https://github.com/pi-hole/docker-pi-hole/actions/runs/7431563473


yubiuser:
Worked only partly.

PromoFaux:
indeed, v annoying. Playing about in a fork!

PR #1576, opened 2024-05-08
Install grep to avoid issues in pihole -w/b with the default busybox …

PromoFaux:

Fixes https://github.com/pi-hole/docker-pi-hole/issues/1577 by replacing `busybox`'s `grep` - which does not include the `P` option

Before:
```text
dev-v6:/# pihole -b test.com
grep: unrecognized option: P
BusyBox v1.36.1 (2023-11-07 18:53:09 UTC) multi-call binary.

Usage: grep [-HhnlLoqvsrRiwFE] [-m N] [-A|B|C N] { PATTERN | -e PATTERN... | -f FILE... } [FILE]...

Search for PATTERN in FILEs (or stdin)

        -H      Add 'filename:' prefix
        -h      Do not add 'filename:' prefix
        -n      Add 'line_no:' prefix
        -l      Show only names of files that match
        -L      Show only names of files that don't match
        -c      Show only count of matching lines
        -o      Show only the matching part of line
        -q      Quiet. Return 0 if PATTERN is found, 1 otherwise
        -v      Select non-matching lines
        -s      Suppress open and read errors
        -r      Recurse
        -R      Recurse and dereference symlinks
        -i      Ignore case
        -w      Match whole words only
        -x      Match whole lines only
        -F      PATTERN is a literal (not regexp)
        -E      PATTERN is an extended regexp
        -m N    Match up to N times per file
        -A N    Print N lines of trailing context
        -B N    Print N lines of leading context
        -C N    Same as '-A N -B N'
        -e PTRN Pattern to match
        -f FILE Read pattern from file
grep: unrecognized option: P
BusyBox v1.36.1 (2023-11-07 18:53:09 UTC) multi-call binary.

Usage: grep [-HhnlLoqvsrRiwFE] [-m N] [-A|B|C N] { PATTERN | -e PATTERN... | -f FILE... } [FILE]...

Search for PATTERN in FILEs (or stdin)

        -H      Add 'filename:' prefix
        -h      Do not add 'filename:' prefix
        -n      Add 'line_no:' prefix
        -l      Show only names of files that match
        -L      Show only names of files that don't match
        -c      Show only count of matching lines
        -o      Show only the matching part of line
        -q      Quiet. Return 0 if PATTERN is found, 1 otherwise
        -v      Select non-matching lines
        -s      Suppress open and read errors
        -r      Recurse
        -R      Recurse and dereference symlinks
        -i      Ignore case
        -w      Match whole words only
        -x      Match whole lines only
        -F      PATTERN is a literal (not regexp)
        -E      PATTERN is an extended regexp
        -m N    Match up to N times per file
        -A N    Print N lines of trailing context
        -B N    Print N lines of leading context
        -C N    Same as '-A N -B N'
        -e PTRN Pattern to match
        -f FILE Read pattern from file
  [✗] test.com is not a valid argument or domain name!
``` 

After:

```bash
dev-v6:/# pihole -b test.com
  [i] Adding test.com to the blacklist...
  [✓] Reloading DNS lists
```

image

---

---

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -21,6 +21,8 @@ RUN apk add --no-cache \
     coreutils \
     curl \
     git \
+    # Install grep to avoid issues in pihole -w/b with the default busybox grep
+    grep \
     iproute2-ss \
     jq \
     libcap \
@@ -34,7 +36,7 @@ RUN apk add --no-cache \
     tini \
     tzdata \
     unzip \
-    wget
+    wget 
 
 ADD https://ftl.pi-hole.net/macvendor.db /macvendor.db
 COPY crontab.txt /crontab.txt


rdwebdesign:
Question:
Why do we need to use the `--upgrade` option?

I tested adding `grep` to the list of packages to be installed and the result was the same.

PromoFaux:
Maybe we do not... @yubiuser made the comment [here](https://discourse.pi-hole.net/t/whitelist-blacklist-options-result-in-grep-errors/65521/3?u=promofaux) that it did not work when he tried, and so I took the `upgrade` option from a comment you made elsewhere and it worked. 

I didn't try it without `upgrade` before based on Yubi's comment, but having since tried it I can confirm it is now working

src/Dockerfile
```diff
@@ -34,7 +34,9 @@ RUN apk add --no-cache \
     tini \
     tzdata \
     unzip \
-    wget
+    wget \
+    # Install grep to avoid issues in pihole -w/b with the default busybox grep
+    grep
```
rdwebdesign, line 39: The other packages are in alphabetical order

rdwebdesign (changes requested):

src/Dockerfile
PromoFaux: Fussy!

rdwebdesign (approved):

yubiuser:
Something must have changed in between so that it is working now with only `grep` being available. 
I testet it with the container now and bosth `-b/-w` work.

P.S. Adding `grep` was the lazy way ;-) `list.sh` still uses the direct database access. At some point we should re-write it to use `FTL` and the new API :-)

PromoFaux:
Yeah, absolutley agree on that - just going for the quickfix option for the time being :)

DL6ER:
We have code for CLI lists via API ready for quite some time but there was an ongoing discussion because this change then introduces authentication for these CLI options while there was none before. We discussed this shortly back in the days but never came to a conclusion what we wanted to do IIRC. Eventually, it was forgotten.

My latest suggestion was to change the default of `webserver.api.localAPIauth` to `false` (it currently default to `true`) but this is also not the optimal solution effectively allowing the CLI to use the API at will without password.

CLI domain searching is also ported to the API and we added a separate option for it (`webserver.api.searchAPIauth`) - if we'd decide for my suggestion, this special option could be removed again - which I'd absolutely love to do!

yubiuser:
I don't remember what my position was when we discussed this back then, but today I think we should not let the cli to bypass the authentication by default. We have https://github.com/pi-hole/pi-hole/blob/development-v6/advanced/Scripts/api.sh which can handle cli authentication so there should be no need to allow bypassing it. 
(I admit it is a bit cumber-stone to re-authenticate for each added black/whitelist entry - unless we store the SID somewhere and don't delete the session)

DL6ER:
Yes, string the session somewhere in the respective user's home should solve this. Maybe we create a directory `~/.pihole` and store a file called `api_sid` in there. We can always throw this against the API and it will either accept or reject it, the latter forcing the user to re-login.

Users will complain, but probably not too hard. We could offer the possibility to create a file `~/.pihole/api_pw` where users can enter either their Pi-hole's normal or application password to cause the CLI to auto-login.

PromoFaux:
Storing the password could work, BUT how would we do that in such a way that it doesn't lead to the possibility of the password being compromised?

yubiuser:
I wouldn't store the password. I think it's fine to request re-login after the session has expired. Saving the SID should make it very convenient already.

DL6ER:
That's what the application password has been designed for. If they use (and store!) it in some Python scripts in their home directory programmatically interacting with the database, why not allow them to store it in an equally "safe" (that's why I said in `/home/<user>`) place?

PromoFaux:
So maybe we just preface it with "you can do x but it is at your own risk"

SID maybe the better option, mind.

DL6ER:
We'll do SID anyway but it will still require relogin after some inactivity.

dschaper:
image

PR #1608, opened 2024-06-30
Default DNS Upstream

PromoFaux:

See https://discourse.pi-hole.net/t/how-to-modify-upstream-server-only-via-gui-on-v6/70943/13?u=promofaux

Lower the barrier to entry/reduce confusion by setting DNS upstreams to `8.8.8.8;8.8.4.4` if we detect that it has not already been set by either the environment variable or config file. Falling back to a default rather than hard-exiting is probably more user friendly, on reflection.

---

---

### DIFF
--- a/src/bash_functions.sh
+++ b/src/bash_functions.sh
@@ -62,12 +62,9 @@ ensure_basic_configuration() {
     fi
 
     # If getFTLConfigValue "dns.upstreams" returns [], exit the container. We need upstream servers to function!
-    if [[ $(getFTLConfigValue "dns.upstreams") == "[]" ]]; then
-        echo ""
-        echo "  [X] No DNS upstream servers are set!"
-        echo "  [i] Recommended: Set the upstream DNS servers in the environment variable FTLCONF_dns_upstreams"
-        echo ""
-        exit 1
+    if [[ $(getFTLConfigValue "dns.upstreams") == "[]" ]]; then                
+        echo "  [i] No DNS upstream set in environment or config file, defaulting to Google DNS"        
+        setFTLConfigValue "dns.upstreams" "[\"8.8.8.8\", \"8.8.4.4\"]"        
     fi
 
     setup_web_password
--- a/src/bash_functions.sh
+++ b/src/bash_functions.sh
@@ -61,7 +61,7 @@ ensure_basic_configuration() {
         chown pihole:pihole /macvendor.db
     fi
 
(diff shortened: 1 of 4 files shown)


yubiuser (commented):
Why not only remove the `exit 1` line and keep the upstream unset + print the warning?

PromoFaux:
Because ultimately nobody will read the warning. 

I originally put the `exit 1` to make it extremely obvious that something needed configuring, but if by defaulting it does not need to be configured immediately.

If we don't default and also don't exit, the startup checks wont always work (e.g in the case of the container having itself as a resolver - either via compose/run or via the host's settings)

test/tests/test_general.py
```diff
@@ -14,14 +14,14 @@ def test_pihole_gid_env_var(docker):
...
-# Finally, tail the FTL log to see if it has been shut down cleanly
+# Finally, grep the FTL log to see if it has been shut down cleanly
 def test_pihole_ftl_clean_shutdown(docker):
     func = docker.run(
         """
         sleep 5
         killall --signal 15 start.sh
         sleep 5
-        tail -f /var/log/pihole-FTL.log
+        grep 'jmpret\|terminated' /var/log/pihole/FTL.log
```
PromoFaux, line 24: Yeah, nice trick here!

Gontier-Julien:
Might i suggest on using maybe Cloudflare Dns (1.1.1.1 & 1.0.0.1) or Quad9 Dns (9.9.9.9 & 149.112.112.112)?

My reasoning behind this is that for Cloudflare is the fastest dns out here, while Quad9 offer better privacy but slower.
Both have a privacy policies meanwhile Google Dns as none as far as i'm aware.

Providing a faster and/or more private dns upstream would be better i think for some default.

Also Cloudflare dns i widely know so i don't think many people would mind + having as faster (even slightly) dns is always a better user experience.

src/bash_functions.sh
```diff
@@ -62,12 +62,9 @@ ensure_basic_configuration() {
     fi
 
     # If getFTLConfigValue "dns.upstreams" returns [], exit the container. We need upstream servers to function!
```
yubiuser, line 64:
```suggestion
    # If getFTLConfigValue "dns.upstreams" returns [], default to Google's DNS server
```

yubiuser (changes requested):

yubiuser (approved):

PR #1609, opened 2024-07-01
Add pytest-clarity to test environment to improve error log output

yubiuser:
We use `pytest` for testing. While it is very capable, its error output is a mess. Adding `pytest-clarity` improves on that (see https://github.com/darrenburns/pytest-clarity).

Before
Screenshot at 2024-07-01 00-10-02

After
Screenshot at 2024-07-01 00-09-46

___

Note: this PR needs https://github.com/pi-hole/docker-pi-hole/pull/1608/commits/99f459e30fe8a9fd8a8b40e99f3a41814d7a1078 to pass the tests

### DIFF
--- a/test/cmd.sh
+++ b/test/cmd.sh
@@ -8,4 +8,4 @@ docker images pihole:${GIT_TAG}
 python -m black ./test/tests/
 
 # TODO: Add junitxml output and have something consume it
-py.test -vv -n auto ./test/tests/
+COLUMNS=120 py.test -vv -n auto ./test/tests/
--- a/test/requirements.txt
+++ b/test/requirements.txt
@@ -2,3 +2,4 @@ pytest == 7.4
 pytest-xdist == 3.5.0
 pytest-testinfra == 10.0.0
 black == 23.12.0
+pytest-clarity == 1.0.1


rdwebdesign:
I really liked the new output, but since the intention is to improve the readability, I think we could use longer lines.

In my opinion, using the default width (80 characters) is creating too many line breaks in the output:

image

We already use line breaks (`/n`) where we need _(and the github interface already prevents horizontal scrollbars)_. 
We could have lines with 120 characters or any other width, adding the `diff-width` parameter [here](https://github.com/pi-hole/docker-pi-hole/blob/e82cfd915cb9282b8f35021e0143fb266a59beee/test/cmd.sh#L11):

```
py.test -vv --diff-width=120 -n auto ./test/tests/
```

What do you think?

PromoFaux:
❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ ❤️ 

But also, what @rdwebdesign said.

PromoFaux:
Although it does not seem to make a difference. 

Still 80 chars with new lines is better than no new lines at all, so I'll take it!

Before:
image


After:
image

rdwebdesign:
Well... the `--diff-width` parameter didn't work as I expected, but I was able to change the width using the `COLUMNS` environment variable:

```
COLUMNS=150 py.test -vv -n auto ./test/tests/
```

image

rdwebdesign (approved):

PR #1621, opened 2024-08-06
Bump alpine to 3.20

PromoFaux:

Update Alpine base from 3.19 to 3.20

---

---

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1
 ARG FTL_SOURCE=remote
-ARG alpine_version="3.19" 
+ARG alpine_version="3.20" 
 FROM alpine:${alpine_version} AS base
 # https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope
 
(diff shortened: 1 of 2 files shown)


yubiuser:
Do you want also upgrade here?
https://github.com/pi-hole/docker-pi-hole/blob/development-v6/test/Dockerfile

PromoFaux:
FP'd

PromoFaux:
Perhaps not as simple as that... one to investigate

rdwebdesign:
I think there is an issue with `python3-dev`.

yubiuser:
Another approach would be to remove the `sed` and the same monkeypatch as used in `core`

https://github.com/pi-hole/pi-hole/blob/be5a8dc51a7b3e7ec3b40a9c815dee1c866ea2d4/test/conftest.py#L21-L35

yubiuser:
@PromoFaux

See my commit and last comment. Which way do you prefer?

PromoFaux:
Ah, Sorry - didn't see that you'd pushed a commit before your comment.

If I've understood it correctly, I think the monkeypatch is probably more resiliant as it is one less thing to have to update every time the python version updates?

yubiuser:
I think I found a nice solution. My concern with the monkeypatch is that it is way harder to understand than a small sed command

yubiuser (approved):

PromoFaux:
Ah yeah, that's good!

PR #1632, opened 2024-09-02
Final changes for merge to `development`

PromoFaux:

---

---


src/Dockerfile
```diff
@@ -5,11 +5,11 @@ FROM alpine:${alpine_version} AS base
 # https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope
 
 ARG TARGETPLATFORM
-ARG WEB_BRANCH="development-v6"
-ARG CORE_BRANCH="development-v6"
-ARG FTL_BRANCH="development-v6"
+ARG WEB_BRANCH="master"
```
yubiuser, line 8: You could also keep `development` here. Then then changes in `test/cmd.sh` would not be necessary....

PromoFaux: That's probably a better idea to be honest.

I'm also thinking I'm going to remove the arm64 tests for now until such time as we can get them working truly isolated from the amd64 tests

yubiuser: > I'm also thinking I'm going to remove the arm64 tests for now until such time as we can get them working truly isolated from the amd64 tests

Why? They are working right now...

PromoFaux: They fail often, but retrying the failed one "fixes" it. 
I think the configs might somehow be leaking between the two arch's tests.

yubiuser (approved):

src/Dockerfile
PromoFaux: image

See...

image


The only way it's possible to fail like that is if it is picking up the config file from the other test container

rdwebdesign: It happened again. The password is already set when the second test runs

yubiuser: Usually they fail for a reason...

PromoFaux: I'd agree if it wasn't that the tests will always pass once I hit retry - something "other" is happening here.. the test is failing because the config file has already been set... nothing in the test sets the config file before the test runs and the only thing I can put it down to is that it's somehow sharing the same volume mount as the other arch tests. 

Actually, I am pretty sure it is. I'm sure it's fixable, but I also don't know how worth it it is.

rdwebdesign: > The only way it's possible to fail like that is if it is picking up the config file from the other test container

The tests are started at the same time.

Is the environment shared between the tests?

PromoFaux: Yeah - I'm pretty sure that's the reason for the failures, but I don't have the capacity to solve it yet, nor do I value the results enough to lose sleep over it today :)

rdwebdesign: Can we run 2 separate tests, using 2 different workflows?

PR #1646, opened 2024-09-13
Tweak README text and formatting

rdwebdesign:
## Description

Mainly format changes, removal of trailing spaces, addition or removal of line breaks and some text corrections.

Using some [markdown alerts](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#alerts) to emphasize critical information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,17 +34,17 @@
 
 > [!NOTE]
 > **Using Watchtower?\
-> See the [Note on Watchtower](#note-on-watchtower) at the bottom of this readme**
+> See the [Note on Watchtower](#note-on-watchtower) at the bottom of this readme**.
 
 > [!TIP]
 > Some users [have reported issues](https://github.com/pi-hole/docker-pi-hole/issues/963#issuecomment-1095602502) with using the `--privileged` flag on `2022.04` and above.\
-> TL;DR, don't use that mode, and be [explicit with the permitted caps](https://github.com/pi-hole/docker-pi-hole#note-on-capabilities) (if needed) instead
+> TL;DR, don't use that mode, and be [explicit with the permitted caps](https://github.com/pi-hole/docker-pi-hole#note-on-capabilities) (if needed) instead.
 
 ## Quick Start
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,37 +2,56 @@
...
 
 ## Upgrade Notes
 
-## !!! THIS VERSION CONTAINS BREAKING CHANGES
-
-### v[ChangeMeBeforeTagging] has been entirely redesigned from the ground up and contains many breaking changes. Environment variable names have changed, script locations may have changed. Please read the the Readme carefully before proceeding
+> [!CAUTION]
+>
+> ## !!! THIS VERSION CONTAINS BREAKING CHANGES
+> **v6 has been entirely redesigned from the ground up and contains many breaking changes.**
```
DL6ER, line 29:
```suggestion
> **v[ChangeMeBeforeTagging] has been entirely redesigned from the ground up and contains many breaking changes.**
```

I suppose `v[ChangeMeBeforeTagging]` is there as Docker is more concerned about tags in the format `YYYY.MM.X`. The tag cannot be foreseen as there is no fixed `v6` release date.

rdwebdesign: I agree we need a better text here, but the current paragraph is not wrong (maybe just incomplete).

`v6` is an entirely redesigned Pi-hole. Every v6 tag will contain breaking changes compared to v5 tags.

rdwebdesign: I changed the text a little bit to warn that replacing older images (`2024.07.0` and earlier) with newer images will cause irreversible changes to config files in the volumes.

I know that docker is related to image tags, but our image tags doesn't contain Pi-hole version number. It is necessary to reinforce that new images will use Pi-hole v6 to avoid confusion.

PromoFaux (approved):

README.md
rdwebdesign: @DL6ER 

After the recent commits, do you still has any suggestions?

Issue #1387, opened 2023-07-19
Logrotate


PromoFaux:
We need a mechanism to ensure logrotate is working

edgd1er:
ideas:
* check if log is listed in var/lib/logrotate/status
* according to rotation (daily, weekly, monthly), log creation date is ok.
* 
https://serverfault.com/questions/189320/how-can-i-monitor-what-logrotate-is-doing

buxm:
So I suppose mine is a known issue...
I've just realised my pihole.log is 8GB and my FTL.log is 70MB.
Both stopped rotating back in August 2022.
I'm running release 2024.02.0 on Synology DSM 7.1.1-42962 Update 6.

Here is what I've tried:
```
#docker exec -it pihole1 bash
root@pihole1:/# ls -lah /var/log/pihole
total 8.1G
drwx------ 1 www-data    999  290 Aug 25  2022 .
drwxr-xr-x 1 root     root    174 Feb 20 22:33 ..
-rw-r--r-- 1 www-data pihole  70M Feb 20 22:42 FTL.log
-rwx------ 1 root     root    14K Aug 25  2022 FTL.log.1
-rwx------ 1 root     root   2.5K Aug 24  2022 FTL.log.2.gz
-rwx------ 1 root     root    20K Aug 23  2022 FTL.log.3.gz
-rwx------ 1 root     root    19M Jul  9  2022 pihole-FTL.log
-rw-r--r-- 1 www-data pihole 8.0G Feb 20 22:49 pihole.log
-rwx------ 1 root     root   9.3M Aug 25  2022 pihole.log.1
-rwx------ 1 root     root   849K Aug 24  2022 pihole.log.2.gz
-rwx------ 1 root     root   703K Aug 23  2022 pihole.log.3.gz
-rwx------ 1 root     root   880K Aug 22  2022 pihole.log.4.gz
-rwx------ 1 www-data root   8.4K Feb 20 22:05 pihole_updateGravity.log
root@pihole1:/# pihole flush once
  [i] Flushing /var/log/pihole/pihole.log ...error: Ignoring /etc/pihole/logrotate because the file owner is wrong (should be root or user with uid 0).
  [✓] Flushed /var/log/pihole/pihole.log
  [✓] Deleted  queries from database
```
I've tried to force root ownership of the log files:
```
root@pihole1:/# chown root:root /var/log/pihole/pihole.log /var/log/pihole/pihole-FTL.log
[...]

PR #1655, opened 2024-10-27
logrotate: fix permissions on configfile

quinte17:
this should fix
https://github.com/pi-hole/docker-pi-hole/issues/1387

## Description
permissions on conifgfile of logrotate are wrong

## Motivation and Context
this was easy to fix

## How Has This Been Tested?
manually in docker container

## Types of changes
installation of logrotate file places the config now in mode 0600

## Checklist:

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -53,7 +53,7 @@ RUN cd /etc/.pihole && \
     install -Dm755 -t /opt/pihole ./advanced/Scripts/*.sh && \
     install -Dm755 -t /opt/pihole ./advanced/Scripts/COL_TABLE && \
     install -Dm755 -d /etc/pihole && \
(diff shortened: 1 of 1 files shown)


yubiuser (commented):
Why do you think the permissions are wrong? We use `644` on bare metal as well.

https://github.com/pi-hole/pi-hole/blob/250fbd217f1c798ab365b37ed5ba7296c5d20f97/advanced/Templates/pihole-FTL-prestart.sh#L19-L21

https://github.com/pi-hole/pi-hole/blob/250fbd217f1c798ab365b37ed5ba7296c5d20f97/automated%20install/basic-install.sh#L1583

quinte17:
i am sry, i did not try to get the error in the first place.
this can be closed.

PR #1687, opened 2025-02-08
Clarify log message for missing caps

jnodorp:
## Description

Changed warning to error.

## Motivation and Context

Confusing container exit without errors.

## Types of changes

## Checklist:

### DIFF
--- a/src/bash_functions.sh
+++ b/src/bash_functions.sh
@@ -235,8 +235,8 @@ fix_capabilities() {
             exit 1
         fi
     else
-        echo "  [!] WARNING: Unable to set capabilities for pihole-FTL."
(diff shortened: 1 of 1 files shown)


yubiuser:
I think I don't understand your motivation. Without the capabilities, FTL would not run anyway - why not stop the container already at startup then?

rdwebdesign:
> Running the container without capabilities is not possible.

This is exactly why we should exit in case the system is unable to set capabilities. The container will fail anyway.

jnodorp:
Sorry, just started out w/ pihole. Then the bug is the warning vs. error message. I updated the PR accordingly.

Having the container exit without an error was confusing.

yubiuser (approved):

PR #1690, opened 2025-02-10
Set default UID and GID to 1000

yubiuser:
Set the default `GID` and `UID` to 1000. Currently, alpine's `busybox` `adduser/addgroup` default to `100/101` which can cause conflicts with file permissions  volumes mounted from the host.

On most (supported) linux systems, the [first user account has a UID of 1000](https://www.wikiwand.com/en/articles/User_identifier#Conventions). If we create files from within the container, they can't be accessed by the host's default user and likely belong to some kind of system account.

The same [holds true for the GID](https://www.wikiwand.com/en/articles/Group_identifier#Conventions).

It's still possible to set `PIHOLE_UID` and `PIHOLE_GID` as env vars to overwrite the IDs


PromoFaux:
I guess my only small concern here is that this change could cause breakages where users have chown'd their host directory mounts to belong to 100/101 - but... it's not really a concern at all. 

@pi-hole/debug for awareness of this change - on the slim chance people's mounts are no longer accessible on the `development` branch.

PromoFaux (approved):

PR #1692, opened 2025-02-18
Development -> Master

PromoFaux:
So that we may release v6 image


yubiuser (approved):
DCO can be overwritten

PromoFaux:
Yeah, not worried about the DCO one for the time being, will bypass the rules

PR #1700, opened 2025-02-18
Add test to output FTL's version information

yubiuser:
Adds a simply test to check for the output of `pihole-FTL -vv`. Main use is probably debugging.


yubiuser:
Second commit adds FTL version output while building the image

PromoFaux (approved):

PR #1719, opened 2025-02-20
Add note about FTLCONF_misc_etc_dnsmasq_d

yubiuser:
Amend the README on `master` to include notes about `FTLCONF_misc_etc_dnsmasq_d`. There are more users using custom configs than we thought.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ services:
     volumes:
       # For persisting Pi-hole's databases and common configuration file
       - './etc-pihole:/etc/pihole'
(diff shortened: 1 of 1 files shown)


arpitgupta:
Can we also add a note about `FTLCONF_misc_dnsmasq_lines` as another option with an example of how to specify the value for array fields? This could help folks avoid having to mount the dnsmasq config.

Also will be useful for these to be in the [migration](https://docs.pi-hole.net/docker/upgrading/v5-v6/) guide?

Furthermore I could not find any mention in the [docs](https://docs.pi-hole.net/docker/configuration/) for how to deal with array fields. Did see it mention in the readme here.

yubiuser:
https://github.com/pi-hole/docs/pull/1173

PromoFaux (approved):

PR #1739, opened 2025-02-22
Allow custom web interface URI for Pi-hole admin

pepramon:
### Overview

This pull request adds support for a configurable web interface URI in the Pi-hole Docker container. With this update, users can specify where the Pi-hole admin interface is served by setting the environment variable FTLCONF_webserver_paths_webhome.

### Motivation

In many deployment scenarios—especially those using a reverse proxy for HTTPS—it is beneficial to serve the admin interface under a custom URI. For example, by setting FTLCONF_webserver_paths_webhome=/pihole/ in your docker-compose file, the admin interface will be available at /pihole/ instead of the default /admin/.

### How to Enable

To activate this feature, add the following line to your docker-compose file:

FTLCONF_webserver_paths_webhome=/pihole/

This instructs the container to serve the admin interface from the specified path.

### Testing

Please test the change in your environment by defining the variable and verifying that the Pi-hole admin interface is correctly accessible from the new URI.

### Conclusion

This update provides enhanced flexibility for configuring the Pi-hole Docker container, making it easier to integrate with custom routing or proxy setups.

### DIFF
--- a/src/bash_functions.sh
+++ b/src/bash_functions.sh
@@ -156,6 +156,8 @@ ftl_config() {
         if [ -d "/var/www/html/admin" ]; then
             echo "Moving /var/www/html/admin to /var/www/html/$WEBHOME"
             mv /var/www/html/admin "/var/www/html/$WEBHOME"
+            # Link to hardcoded scripts detects path
+            ln -s "/var/www/html/$WEBHOME" /var/www/html/admin
         fi
     fi
 
--- a/src/bash_functions.sh
+++ b/src/bash_functions.sh
@@ -145,6 +145,19 @@ ftl_config() {
         echo "  [i] No DNS upstream set in environment or config file, defaulting to Google DNS"
         setFTLConfigValue "dns.upstreams" "[\"8.8.8.8\", \"8.8.4.4\"]"
     fi
+    
+    # If the function getFTLConfigValue("webserver.paths.webhome") returns a value different from "/admin/",
+    # Then, move the directory "/var/www/html/admin" to the new location.
+    if WEBHOME=$(getFTLConfigValue "webserver.paths.webhome") && [[ "$WEBHOME" != "/admin/" ]]; then
+        # Remove leading slash if present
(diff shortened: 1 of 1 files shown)


yubiuser:
This might work in docker already, but on bare metal this won't work as we (unfortunately) have some hard-coded paths in various scripts.

rdwebdesign:
I don't think this will work in docker without other modifications.

rdwebdesign:
> I do not know how `pihole updatechecker` works

It uses git to compare the local and remote branches.

These lines won't work if the web interface is moved:

https://github.com/pi-hole/pi-hole/blob/567bb724b18bf174f033d0b8b289cc44a5e9c8eb/advanced/Scripts/updatecheck.sh#L88-L95

Without updatechecker, the container will fail to find the current local version.

rdwebdesign:
I really think this should be fixed on Core as well before any release, so every installation (docker and bare metal) will allow this option.

And when we change Core code, some changes made here will need to be adjusted, some will be unnecessary.

PromoFaux:
No comment on inclusion validity at the moment, but from a technical standpoint:

The login page goes brrrrrrrrrrrrrrrrrrrrrrr. Which it probably shouldn't 😉 

loginpagebrrrr

rdwebdesign:
Looks like the page is redirecting to itself (`/admin/login`) or constantly reloading.

PromoFaux:
> I really think this should be fixed on Core as well before any release,

There are also some hardcoded `admin/` paths in the web interface by the looks of things

image

rdwebdesign:
And in FTL too:
https://github.com/pi-hole/FTL/blob/ac500d5f1ff192d087209d6e9b955515c8f35434/src/api/docs/content/index.html

pepramon:
Oups, I do not detect that issue, my pihole is behind a nginx proxy, and seem that happens with SSL fail. Now I see 

Later, when you enter (jut try to write and stops), seems that all works nice. It's only the login page.

I think that better option is remove this pull request because is harder than I think.

Sorry.

PR #1741, opened 2025-02-22
Include FTLCONF_dns_listeningMode: 'all' in the compose example file

yubiuser:
Makes it more visible in the example file that `FTLCONF_dns_listeningMode: 'all'` usually is the default in dockers bridge mode.
Brings the example in line with https://docs.pi-hole.net/docker/

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,8 @@ services:
       TZ: 'Europe/London'
       # Set a password to access the web interface. Not setting one will result in a random password being assigned
       FTLCONF_webserver_api_password: 'correct horse battery staple'
+      # If using Docker's default `bridge` network setting the dns listening mode should be set to 'all'
+      FTLCONF_dns_listeningMode: 'all'
     # Volumes store your data between container upgrades
     volumes:
       # For persisting Pi-hole's databases and common configuration file
@@ -92,7 +94,6 @@ services:
 ```
 
 2. Run `docker compose up -d` to build and start pi-hole (Syntax may be `docker-compose` on older systems).
-3. If using Docker's default `bridge` network setting, set the environment variable `FTLCONF_dns_listeningMode` to `all`.
 
 > [!NOTE]
 > Volumes are recommended for persisting data across container re-creations for updating images.


README.md
```diff
@@ -74,6 +74,8 @@ services:
       TZ: 'Europe/London'
       # Set a password to access the web interface. Not setting one will result in a random password being assigned
       FTLCONF_webserver_api_password: 'correct horse battery staple'
+      # If using Docker's default `bridge` network setting the dns listening mode should be set to 'all'
+      FTLCONF_dns_listeningMode: 'all'
```
rdwebdesign, line 78: Do you think this should be added as a comment in the example?
```suggestion
      # FTLCONF_dns_listeningMode: 'all'
```

PromoFaux:
Is it case sensitve? 

image

yubiuser:
Maybe, but this will be fixed soon.

https://github.com/pi-hole/FTL/pull/2256

PromoFaux:
I meant the value rather than the key :)

yubiuser:
Mhhh... this could explain https://github.com/pi-hole/docker-pi-hole/issues/1744

yubiuser:
It's here, and the function is case insensitive

https://github.com/pi-hole/FTL/blob/ac500d5f1ff192d087209d6e9b955515c8f35434/src/datastructure.c#L1319-L1331

https://linux.die.net/man/3/strcasecmp

yubiuser:
Pinging @DL6ER 

Are the values case sensitive or not? [This issue](https://github.com/pi-hole/docker-pi-hole/issues/1744) suggest so, but the `strcasecmp` should not care.

DL6ER:
They are not and I have also not been able to reproduce this?

```
$ grep "listening" /etc/systemd/system/pihole-FTL.service
Environment=FTLCONF_dns_listeningMode='aLL'
```

```
2025-02-24 19:00:12.267 CET [2021954M] INFO: 1 FTLCONF environment variable found (1 used, 0 invalid, 0 ignored)
2025-02-24 19:00:12.267 CET [2021954M] INFO:    [✓] FTLCONF_dns_listeningMode is used
```

yubiuser:
Turned out, the OP from the [linked issue](https://github.com/pi-hole/docker-pi-hole/issues/1744#issuecomment-2677098110) was also not able to reproduce it in the end.

yubiuser:
After we sorted it, I think this is good to go.

PromoFaux (approved):

Issue #1745, opened 2025-02-23
Setting FTLCONF_dns_port to something other than 53 causes the health check to fail


nwctenor:
When setting the FTLCONF_dns_port environment variable, the container's healthcheck continues to run the _dig_ command without changing the port to the new port number.

This is a: **Bug**  

## Details

In the _environment_ section of my docker-compose.yml I have set "FTLCONF_dns_port: 5353".  The healthcheck runs _HEALTHCHECK CMD dig +short +norecurse +retry=0 @127.0.0.1 pi.hole || exit 1_ which does not include the new port that FTL is running on. (This is based on the Dockerfile)

## Related Issues
- [ ] I have searched this repository/Pi-hole forums for existing issues and pull requests that look similar 

## How to reproduce the issue 

1. Environment data
  * Operating System: **Ubuntu** 
  * Hardware: Virtual instance on Vultr  
  * Kernel Architecture: x86_64 
  * Docker Install Info and version: 
    - Software source: OS provided package
    - Supplimentary Software: 
  * Hardware architecture: x86_64 

2. docker-compose.yml contents, docker run shell command, or paste a screenshot of any UI based configuration of containers here
name: dns
services:
  pihole:
    depends_on: 
      - dnscrypt
    env_file: "./.env"
    container_name: pihole
    hostname: pihole
    domainname: ##DOMAINNAME##
    image: pihole/pihole:latest
    ports:
      - "${WGSRV0_IPV4}:53:5353/tcp"
      - "${WGC0_IPV4}:53:5353/tcp"
      - "${WGSRV0_IPV4}:53:5353/udp"
      - "${WGC0_IPV4}:53:5353/udp"
      - "${WGC0_IPV4}:80:8080/tcp"
      - "${WGC0_IPV4}:443:8443/tcp"
    dns: 172.28.0.2
    dns_search:
      - ##SEARCH1##
      - ##SEARCH2##
    environment:
      TZ: 'America/Chicago'
      FTLCONF_webserver_api_password: '##PASSWORD##'
      FTLCONF_dns_listeningMode: 'all'
      FTLCONF_dns_upstreams: dnscrypt#5353
      FTLCONF_webserver_port: "8080r,8443s,[::]:8080ro,[::]:8443so"
      FTLCONF_dns_port: 5353
      #FTLCONF_dns.reply.host.IPv4: "10.9.0.2"
      FTLCONF_webserver_domain: "##DOMAIN##"
    volumes:
[...]

PR #1746, opened 2025-02-23
Honor configured DNS port for docker health check

yubiuser:
Fixes https://github.com/pi-hole/docker-pi-hole/issues/1745

When users configure an alternative DNS listening port, the container health check should take this into account.


FibreFoX:
Shouldn't this be read from the `pihole.toml` instead? ENV can be missing and the healthcheck still would be wrong.

yubiuser:
Excellent idea

rdwebdesign (approved):

PR #1758, opened 2025-02-25
WEBPASSWORD_FILE is mounted to /run/secrets

yubiuser:
Build on top of https://github.com/pi-hole/docker-pi-hole/pull/1750

If using secrets with Docker swarm/compose, they are usually referenced by the file path **on the host** and mounted by  default to `/run/secret/my_secret`. 
This PR amends the Readme and changes the check to the appropriate location within the container.

https://docs.docker.com/compose/how-tos/use-secrets/#use-secrets
https://docs.docker.com/engine/swarm/secrets/#how-docker-manages-secrets

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ To explicitly set no password, set `FTLCONF_webserver_api_password: ''`.
 | `FTLCONF_[SETTING]` | unset | As per documentation | Customize pihole.toml with settings described in the [API Documentation](https://docs.pi-hole.net/api).<br><br>Replace `.` with `_`, e.g for `dns.dnssec=true` use `FTLCONF_dns_dnssec: 'true'`.<br/>Array type configs should be delimited with `;`.|
 | `PIHOLE_UID` | `1000` | Number | Overrides image's default pihole user id to match a host user id.<br/>**IMPORTANT**: id must not already be in use inside the container!|
 | `PIHOLE_GID` | `1000` | Number | Overrides image's default pihole group id to match a host group id.<br/>**IMPORTANT**: id must not already be in use inside the container!|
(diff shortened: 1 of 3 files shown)


yubiuser:
This is also how other users understand the handling of the `secret` function.

https://github.com/pi-hole/docs/pull/1182

yubiuser:
We could also go a step further to minimize duplication and require a fixed secret_name: Currently one needs to do

```
services:
  pihole:
    container_name: pihole
    environment:
      WEBPASSWORD_FILE: 'pihole_webpasswd'
    secrets:
      - pihole_webpasswd

secrets:
  pihole_webpasswd:
    file: ./pihole_password.txt
```

Which is kind of duplicating the secret_name into `WEBPASSWORD_FILE`
An other approach would be to set a fixed 'name' for the secret, e.g. 'pihole_webpasswd' and skip the check for `${WEBPASSWORD_FILE}' in the script and go straight to check [ -r "/run/secrets/pihole_webpasswd" ]

buckaroogeek:
> An other approach would be to set a fixed 'name' for the secret, e.g. `pihole_webpasswd` and skip the check for `${WEBPASSWORD_FILE}` in the script and go straight to check `[ -r "/run/secrets/pihole_webpasswd" ]`

This approach does offer a little more simplicity which can always be good, especially for those looking for cookbook to copy and paste.

A combination approach when such that using docker secrets, check for a default name and use that if present, otherwise use a value set in WEBPASSWORD_FILE? This does add more complexity to the code but pehaps some simplicity to users? This would be in line with both the Simple and Advanced examples in the Docker Compose documentation.

PromoFaux:
Yep - OK, I like it

Issue #1752, opened 2025-02-24
S6 no longer in image


PromoFaux:
> Hey, i am not sure if this is exactly my issue, but i contributed to a separate docker image based on the official pihole:latest image and am now trying to update it to the latest changed. In the past the image used `/etc/services.d/` to start other services separately from running the main pihole entrypoint without overwriting anything. This was a very good and working approach, but now i cannot use this anymore, because it seems like the alpine image does not contain any kind of init system.
> I tried changing the entrypoint (which i really do not want to do, as i do not want to interferre with the entrypoint of the base pihole image) but that did not work either, i kept getting errors, that it couldnt start and stopped the container.
> 
> Therefore could you please add an init system OR add a mechanic do expand the startup of the image maybe through additional scripts in a specific directory. 

 _Originally posted by @Joly0 in [#1701](https://github.com/pi-hole/docker-pi-hole/issues/1701#issuecomment-2677170049)_

PromoFaux:
@Joly0 - created a new issue because the discussion is not on topic to the issue you tacked onto .

This is not likely to happen, I'm afraid. V6 was a chance to redesign the image from the ground up, and as we no longer need to run multiple binaries, we no longer needed the s6 init system. This also is more in fitting with the "docker way" of doing things, which is one service per container.

Take a look at projects such as this one https://github.com/klutchell/balena-pihole, where additional services are added in a stack, rather than injecting them into a single monolithic image.

Joly0:
Hey @PromoFaux thank your for creating a separate issue. I have looked through various github issues to see, how others are creating their adjusted image and most of them (if not all of them now) create a separate wrapper script, that overwrites the default start.sh script, executes the custom stuff and in the end usually executes the start.sh, this is what balena-pihole does.

While all in all this is a solution one could use and is working, i dislike this approach, because you have to keep track of the upstream (meaning the repo here pi-hole/docker-pi-hole) and check regularly if the entrypoint has changed, which is not super likely but can happen and would probably break the container one created.

I do understand the issue regarding one service per container and that this project no longer needs an init system, but it makes it harder to built on top of it and create separate images based on your great work here. Also, and this is only my personal opinion, an init system does not violate the "one service per container" rule, because it is not really a service itself, but part of another service. The init system is only needed to start and gracefully stop the actual services in a way, so no zombie processes are left behind. But of course i understand, that an init system would require a slightly larger image from you and would require some more work to keep the init system itself up to date.

[...]

Issue #1701, opened 2025-02-18
Removal of `/s6-init` breaks `entrypoint` overrides


troykelly:
### Versions

Core version is v6.0 (Latest: v6.0)
Web version is v6.0 (Latest: v6.0)
FTL version is v6.0 (Latest: v6.0)

### Platform

- OS and version:  Debian GNU/Linux 11

- Platform:  Docker

### Expected behavior

Anybody with old entrypoint overrides will be calling `/s6-init` as the last command to allow the container to continue to start as intended, ie:

```yaml
    entrypoint: |
      /bin/sh -c 
      "echo 'server=/61.10.in-addr.arpa./127.0.0.1#953' > /etc/dnsmasq.d/02-custom.conf &&
       echo 'server=/an.example.invalid./127.0.0.1#953' >> /etc/dnsmasq.d/02-custom.conf && 
       echo 'server=/another.example.invalid./127.0.0.1#953' >> /etc/dnsmasq.d/02-custom.conf && 
       echo 'server=/yet.another.example.invalid./127.0.0.1#953' >> /etc/dnsmasq.d/02-custom.conf && 
       exec /s6-init"
```

### Actual behavior / bug

Without the former startup file, pihole keeps restarting.

### Steps to reproduce

1. Override entrypoint on container
2. Start container

## Debug Token

- URL:  N/A

### Screenshots

_If applicable, add screenshots to help explain your problem._

### Additional context

I can't see anything announced that the startup entrypoint was changing, but using the new entrypoint resolves the issue.

To resolve this issue, use `start.sh`

```yaml
    entrypoint: |
      /bin/sh -c 
      "echo 'server=/61.10.in-addr.arpa./127.0.0.1#953' > /etc/dnsmasq.d/02-custom.conf &&
       echo 'server=/an.example.invalid./127.0.0.1#953' >> /etc/dnsmasq.d/02-custom.conf && 
       echo 'server=/another.example.invalid./127.0.0.1#953' >> /etc/dnsmasq.d/02-custom.conf && 
       echo 'server=/yet.another.example.invalid./127.0.0.1#953' >> /etc/dnsmasq.d/02-custom.conf && 
       exec /usr/bin/start.sh"
```

Yes, this isn't the greatest hack - but there's no way to persist minor config settings like this without shared volumes, so creating this file on startup is the only real solution.

PromoFaux:
[...]

PR #1773, opened 2025-03-03
Added functionality to expand startup of container with custom scripts

Joly0:
`{Please select 'base: development' as target branch above! (you can delete this line)}`

## Description

Added functionality to the start.sh script so custom scripts can be mounted and started from the "/custom-cont-init.d" directory inside the docker container. All scripts in the directory are `chmod +x`´ed and then executed in sorted order, but in background and "only" with a 10 seconds sleep timer between each script execution. So if scripts are dependant on each other, one would need a mechanic to handle that.

As an example i created 3 test scripts:
- 10-successful-script.sh
- 20-failing-script.sh
- 30-endless-script.sh

These are executed in order with a 10 second delay in between but if the first script would take longer to execute, the second script would start nontheless. So each script does neither interfere with other scripts from that folder, nor do they interfere with the way the pihole processes are started, as everything is run in background/parallel. Also output from those scripts should be captured correctly and forwarded so they can be viewed from the docker logs, though i am not sure if processes that are started in background from one of those scripts are handled in the same way. Thats something i cant really tell for sure.

## Motivation and Context

This PR solves https://github.com/pi-hole/docker-pi-hole/issues/1752 
With this PR the pihole docker image can be used as a base image for other containers so other containers can have their own startup scripts without interfering with the way pihole starts.
Also this PR adds a mechanic for users to simply mount bash scripts into the container, that are executed at runtime for whatever reason, might solve https://github.com/pi-hole/docker-pi-hole/issues/1701#issuecomment-2677170049

## How Has This Been Tested?

I have created the above mentioned scripts with this code:
10-successful-script.sh:
```
#!/bin/bash

# Test script that will always succeed
echo "[SUCCESS] Custom init script 10-successful-script.sh is running"

# Create a test file to verify the script ran
touch /tmp/custom-init-success.txt
echo "Script executed at $(date)" > /tmp/custom-init-success.txt

# Simulate some work with output to Docker logs
for i in {1..5}; do
  echo "[SUCCESS] Working... step $i of 5"
  sleep 1
done

echo "[SUCCESS] Custom init script completed successfully"
echo "[SUCCESS] Also wrote to /tmp/custom-init-success.txt for verification"

# Exit with success
exit 0
```
20-failing-script.sh:
```
#!/bin/bash

# Test script that will fail
echo "[FAILURE TEST] Custom init script 20-failing-script.sh is running"

# Create a test file to verify the script ran but failed
touch /tmp/custom-init-failure.txt
echo "Script started at $(date)" > /tmp/custom-init-failure.txt

# Simulate some work before failing
echo "[FAILURE TEST] Doing some work before failure"
sleep 2

# Try to create a file in a location that doesn't exist (this will cause an error)
echo "[FAILURE TEST] About to fail..."
mkdir -p /nonexistent_directory_that_requires_root_permissions/test 2>&1

# This command will generate an error
echo "This should not appear in the success results" > /nonexistent_directory_that_requires_root_permissions/test/file.txt 2>&1

# Record failure in our log file and Docker logs
echo "[FAILURE TEST] Script failed with exit code $?" | tee -a /tmp/custom-init-failure.txt

# Exit with error
exit 1
```
30-endless-script.sh:
```
#!/bin/bash

# Configuration
UPDATE_INTERVAL=30  # Output status every 30 seconds

# Initialize with startup message
echo "[ENDLESS] Script started at $(date)"
echo "[ENDLESS] This script will run indefinitely and log status every $UPDATE_INTERVAL seconds"
echo "[ENDLESS] Monitor the output in docker logs"

# Counter for iterations
counter=0

# Create a function to handle cleanup when the script is terminated
cleanup() {
  echo "[ENDLESS] Script terminating at $(date) - received signal"
  exit 0
}

# Set up signal handling
trap cleanup SIGTERM SIGINT

# Endless loop
while true; do
  # Increment counter
  ((counter++))
  
  # Log current status to Docker logs
  echo "[ENDLESS] Still running - iteration $counter at $(date)"
  echo "[ENDLESS] Memory usage: $(free -m | grep Mem | awk '{print $3}')MB used"
  echo "[ENDLESS] System load: $(cat /proc/loadavg)"
  echo "-------------------------------------------"
  
  # Sleep for the specified interval
  sleep $UPDATE_INTERVAL
done
```

The first script just runs once and exits without an error, the second one exits with an error code of 1 and the third script keeps running endlessly and continuesly outputs to the stdout which is visible in the docker logs.

All these scripts can be placed in the current working directory and the docker container can be started with a command similar to this:
`docker run -d --name pihole-test -v "$(pwd):/custom-cont-init.d" pihole/pihole:latest`
Afterwards viewing the docker logs should show the output of the different scripts that have been mounted.

## Types of changes

## Checklist:

### DIFF
--- a/src/bash_functions.sh
+++ b/src/bash_functions.sh
@@ -258,3 +258,49 @@ fix_capabilities() {
     fi
     echo ""
 }
+
+# New function to run custom initialization scripts
+run_custom_init_scripts() {
+  local custom_init_dir="/custom-cont-init.d"
+  
+  # Check if the directory exists
+  if [ ! -d "$custom_init_dir" ]; then
+    echo "  [i] Custom init directory $custom_init_dir does not exist, skipping"
+    return 0
+  fi
+  
+  echo "  [i] Checking for custom initialization scripts in $custom_init_dir"
+  
+  # Find all executable scripts or make them executable
+  find "$custom_init_dir" -type f | sort | while read -r script; do
+    # Check if file is already executable
+    if [ ! -x "$script" ]; then
+      echo "  [i] Making $script executable"
+      chmod +x "$script" 2>/dev/null || {
+        echo "  [!] Failed to make $script executable, skipping"
+        continue
+      }
+    fi
+    
+    # Extract the filename
+    filename=$(basename "$script")
+    
+    # Check if the filename starts with a number
+    if [[ "$filename" =~ ^[0-9]+ ]]; then
+      echo "  [i] Running custom init script: $script"
+      # Run the script in background with output redirected to docker logs
+      # Use 'exec' to run in a subshell and avoid blocking
+      sleep 10
+      (exec "$script" 2>&1 | sed "s/^/[Custom Init: $filename] /") &
+      
+      # Store the PID in case we want to track it later
+      local script_pid=$!
+      echo "  [i] Started $script with PID $script_pid"
+    else
+      echo "  [i] Skipping $script, filename doesn't start with a number"
+    fi
+  done
+  
+  echo "  [i] Finished processing custom initialization scripts"
+  return 0
+}
--- a/src/start.sh
+++ b/src/start.sh
@@ -45,6 +45,9 @@ start() {
   # Install the logrotate config file
   install_logrotate
 
+  # Run custom initialization scripts (new addition)
+  run_custom_init_scripts || echo "  [!] Error running custom init scripts, continuing with startup"
+
   #migrate Gravity Database if needed:
   migrate_gravity
 


PromoFaux:
Thanks for the submission. You've managed to include some additional unrelated commits from the master branch here, but not to worry - I'll review based on the code and if we decide to go this way we can cherrypick.

I'll be honest, I'm still in two minds about the inclusion of this. What concerns me the most is what becomes _effectively_ the ability for arbitrary scripts to be injected into the container _and_ run as root.

Joly0:
> Thanks for the submission. You've managed to include some additional unrelated commits from the master branch here, but not to worry - I'll review based on the code and if we decide to go this way we can cherrypick.
> 
> I'll be honest, I'm still in two minds about the inclusion of this. What concerns me the most is what becomes _effectively_ the ability for arbitrary scripts to be injected into the container _and_ run as root.

Ok, thank you for your feedback. I also noticed the unrelated commits, but i made my pushes against master and not against development, i think thats why there are additional commits in my PR.

Regarding your concern i think this is a non-issue. The user would have to mount the scripts folder on the host and someone would need to have access to that mounted folder on the host and the unlikely event, that the container is restarted in a near future. While i see your point, if someone has that level of access to place any files in the mounted folder and restart the container, then that person would most probably have access to the host system itself and could do whatever they want. If pihole itself had a security problem and someone gained access to the pihole container that wouldnt be an issue aswell, because pihole is running as a non root user, which means the user would be a non root user aswell, limiting access to the scripts folder.
I might be overlooking something and maybe you have something that i do not currently have in mind, but this is my opinion on this.

PromoFaux:
Closing out - will continue on the original issue thread, but short version: "Not at this time"

Issue #1767, opened 2025-02-28
Fix gravity.sh -- Alpine / Docker specific issue


karolkania:
### Versions

- Core version is v6.0.4 (Latest: v6.0.4)
- Web version is v6.0.1 (Latest: v6.0.1)
- FTL version is v6.0.2 (Latest: v6.0.3)

### Platform

- OS and version:  Alpine

- Platform:  Docker

### Expected behavior

Update Gravity (list of blocked domains) should update the database

### Actual behavior / bug

Update Gravity (list of blocked domains) does not happen

### Steps to reproduce

Add a proper list and go to /admin/gravity; click update.

Output:

```
  Gravity temporary directory does not exist or is not a writeable directory, falling back to /tmp. 
  [✓] DNS resolution is available

  [i] Neutrino emissions detected...

  [✓] Preparing new gravity database
  [✓] Creating new gravity databases
  [✓] Pulling blocklist source list into range
  [i] Using libz compression


  *** HERE's the bug ***
  [✗] Unable to write to /etc/pihole/listsCache
      Please run pihole -g as root

  [✓] Building tree
  [i] Number of gravity domains: 0 (0 unique domains)
  [i] Number of exact denied domains: 0
  [i] Number of regex denied filters: 0
  [i] Number of exact allowed domains: 0
  [i] Number of regex allowed filters: 0
  [✓] Optimizing database
  [✓] Swapping databases
  [✓] The old database remains available
  [✓] Cleaning up stray matter

  [✓] Done.


```

---

Not digging too deep, I was able to figure out that this is most likely related to differences between how the shell script is interpreted by Bash - on Alpine (Ash, not Bash) this seems to trigger unexpected results.

I've raised a quick PR -- https://github.com/pi-hole/pi-hole/pull/6030/commits/a9650ae28736aedd62680f4f705525021185d519

and this is the script output after the fix is applied:

```
Gravity temporary directory does not exist or is not a writeable directory, falling back to /tmp. 
  [✓] DNS resolution is available

  [i] Neutrino emissions detected...

  [✓] Preparing new gravity database
  [✓] Creating new gravity databases
  [✓] Pulling blocklist source list into range
  [i] Using libz compression

directory_permissions: 775
saveLocation_permissions: 660
  [i] Target: https://raw.githubusercontent.com/StevenBlack/hosts/master/hosts
  [✓] Status: No changes detected
[...]

PR #1786, opened 2025-03-14
fix: alpine bash issue

LizenzFass78851:
## Description

This creates a symlink for the `test` and `[` commands to make them executable via `busybox` and not `coreutils`. It writes commands to disable the `bash` internal `test` and `[` commands both in the `bashrc` (for interactive sessions) and in a separate sh file that is used as a script for non-interactive sessions (specified via an env). This way, at least the execution of `bash` internal commands like `test` and `[` can be prevented and these `busybox` commands can be used instead.

## Motivation and Context

This (hopefully) fixes: #1767

## How Has This Been Tested?

I tested this on an Ubuntu 20.04 VM with an outdated containerd.io package to be able to reproduce the bash alpine issue and test its fix. More details [here](https://github.com/pi-hole/docker-pi-hole/issues/1767#issuecomment-2718596501).

## Types of changes

## Checklist:

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -43,13 +43,13 @@ RUN apk add --no-cache \
 
 # Creates symbolic links for specific Bash commands that point to Busybox,
 # and adds these commands to the Bash configuration to avoid compatibility issues on Alpine Linux
-ARG BASH_COMMANDS="test ["
-RUN for CMD in ${BASH_COMMANDS}; do \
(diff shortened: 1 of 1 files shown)


yubiuser:
Thanks for this. I'm currently unable to test this (vacation 🎉) but this is something I had in mind.

If this works as intended, I would also revert https://github.com/pi-hole/pi-hole/pull/6030 as this is more a workout then the proper fix and has it's own regressions.

PR #1788, opened 2025-03-15
Set day of week for gravity update to be the same as bare metal install

mwoolweaver:
## Description

Set day of week for gravity update to be the same as bare metal install

## Motivation and Context

Consistency between docker and bare metal 

Should avoid [this](https://www.reddit.com/r/pihole/s/TguKhvt5yh)

## Types of changes

## Checklist:

- [*] My code follows the code style of this project.


yubiuser (approved):
https://github.com/pi-hole/pi-hole/blob/9fe687bd62a7adb9b7f5f56b58e2823ee58d8fbc/advanced/Templates/pihole.cron#L21

rdwebdesign:
Just thinking:

I'm not sure if the difference was a mistake or intended to spread/split the requests in 2 different days.

yubiuser:
I've been thinking about this too, but both on docker and bare metal we randomize the download times per user to reduce the load on the hosting servers. I don't think it is necessary to distribute across different days.

rdwebdesign (approved):

PR #1820, opened 2025-05-03
Use CODEOWNERS instead of deprecated dependbot/reviewers

yubiuser:
**What does this PR aim to accomplish?:**

Dependabot will drop the 'reviewer' option. One should use CODEOWNER file instead.
See https://github.blog/changelog/2025-04-29-dependabot-reviewers-configuration-option-being-replaced-by-code-owners/


PromoFaux (approved):

yubiuser:
Sure they could. But we also accumulated a lot of fixes in the components repos - we might want to release soon anyway?

PR #1832, opened 2025-06-06
fix: docker base version detection for dependabot

LizenzFass78851:
## Description

- This changes the from lines of the two dockerfiles (src and test) so that depentabot can automatically detect updates to the base images again.
- At the same time, it adds the ability to search for updates of the base image for the docker image located in the test directory

## Motivation and Context

Because otherwise the Docker base updates would be forget. For example, Alpine version 3.22 was released recently. According to the GitHub runner's log, dependabot always runs on Sundays, but it didn't recognize the base image tags and therefore can't searched for updates, for example, Alpine releases.

## How Has This Been Tested?

I tested these changes in my fork and manually triggered dependabot and it recognized the updates for the base images after the change. (I just can't test it more precisely because certain github actions on my account aren't working properly at the moment)

## Types of changes

## Checklist:


PromoFaux:
Thanks!

The only issue I have with this currently is that we set the alpine version in GHA (granted, this is also out of date)

https://github.com/pi-hole/docker-pi-hole/blob/152dd7a3b1340c24f14c59862c291f884f6131c3/.github/workflows/build-and-publish.yml#L24-L27

Which now leaves me wondering what is best to do for the RISCV arch, which needs to be built on `edge` (I think... something may have changed here since added RISC support)

yubiuser:
Do we need some dependabot policy? Because the `cli` image changes quite often.

LizenzFass78851:
@PromoFaux 
I have changed something in the form lines which means that you can assign separate tags for each architecture (e.g. for riscv alpine edge and for the rest alpine stable version). This would also allow dependabot to continue to find alpine updates for the dockerfile.

@yubiuser 
I removed the commit that included the test directory in the dependabot config.

src/Dockerfile
```diff
@@ -101,8 +110,7 @@ RUN if   [ "$TARGETPLATFORM" = "linux/amd64" ];    then FTLARCH=amd64; \
     elif [ "$TARGETPLATFORM" = "linux/arm/v7" ];   then FTLARCH=armv7; \
     # Note for the future, "linux/arm6/v8" is not a valid value for TARGETPLATFORM, despite the CI platform name being that.
     elif [ "$TARGETPLATFORM" = "linux/arm64" ];    then FTLARCH=arm64; \
-    # RISCV is currently not working on GHA, and so FTL is unavailable for this platform.
-    #elif [ "$TARGETPLATFORM" = "linux/riscv64" ];  then FTLARCH=riscv64; \
+    elif [ "$TARGETPLATFORM" = "linux/riscv64" ];  then FTLARCH=riscv64; \
```
PromoFaux, line 113: Good catch... I forgot about this one when I added the risc build back in.

Goes to show how many people are actually using it.

PromoFaux (approved):

yubiuser:
Did work!

https://github.com/pi-hole/docker-pi-hole/pull/1833

PR #1842, opened 2025-06-08
Remove composite login action

yubiuser:
Remove composite login action. The action was used to reduced code duplication, but I think the amount of complexity added does not outweigh the small amount of code reduction.


PromoFaux:
👍 Fair comment

PR #1856, opened 2025-07-01
Build all containers from a fixed base image

yubiuser:
Currently, all images were build on `alpine:3.22` except `riscv`which was based on `edge`. This distinction was introduced by https://github.com/pi-hole/docker-pi-hole/pull/1511 as of then there was only one image type (`edge`) for `riscv`. But not tagged images exist also for `riscv` (see https://hub.docker.com/r/riscv64/alpine/tags). 
Using `edge`when it's not needed can cause issues as it's considered unstable.

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -1,17 +1,7 @@
 # syntax=docker/dockerfile:1
 ARG FTL_SOURCE=remote
 # Pull Stable images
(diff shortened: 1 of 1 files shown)


PromoFaux (approved):
Cool, yep as long as there are real tags now for RISCV - then lets switch to them. Makes the dockerfile less complex!

PR #1843, opened 2025-06-09
Build ARM on ARM and add ARM back to test suite

yubiuser:
- Speeds up building of the images by running ARM-builds on ARM-runners
- Add back tests for ARM architecture (we publish those images, we should test them).
Note. there are no docker-cli images for i386 and RISCV

PR #1859, opened 2025-07-02
Major overhaul to the test suite

yubiuser:
1. Builds `arm` and `riscv` images on arm as it is faster
2. Completely restructure  the test suite
 a) so far, it was a convoluted mess from bash scripts building a python test image which used bash scripts to build the docker-image within itself and run the test in this docker-in-docker container
b) now we use `tox` to build the docker-container and run the tests locally/CI
c) we have support for **all** platforms `docker` supports
d) much less code and less convoluted
e) the pytest's `conftest.py` was cleaned up a lot
f) one major cause for the random failures we've seen previously on the emulated platforms was due to using `pytest-xdist` and running the tests in parallel  - this is removed now
g) another issue I observed was a timing issues, esp. on `RISCV`. Adding some sleep commands before running the test commands fixed those.
h) I added a test that compares the value of the `CIPLATFORM` var to the arch reported by FTL

Supersedes https://github.com/pi-hole/docker-pi-hole/pull/1843


PromoFaux:
Not sure if you saw my comment on mattermost - but could you look at includiung a `setup.py` too? Without it tox refused to run locally...

yubiuser:
Did not check your comment yet.

> Without it tox refused to run locally...

Interesting, runs fine here...

PromoFaux:
```
ERROR: No pyproject.toml or setup.py file found. The expected locations are:
  /home/adam/repos/docker-pi-hole/test/pyproject.toml or /home/adam/repos/docker-pi-hole/test/setup.py
You can
  1. Create one:
     https://tox.readthedocs.io/en/latest/example/package.html
  2. Configure tox to avoid running sdist:
     https://tox.readthedocs.io/en/latest/example/general.html
  3. Configure tox to use an isolated_build
```

yubiuser:
We figured out, that the requirement of `setup.py` is due to an old/outdated `tox` version.

yubiuser:
Just a rebase, no functional change.

PromoFaux:
I need to change the required tests

PromoFaux (approved):

PR #1660, opened 2024-11-16
Updated the instructions to free up port 53 in Fedora

OptimoSupreme:
## Description
1. Added a dedicated Fedora section below Ubuntu's since the Ubuntu instructions no longer work on Fedora.
2. Added 'Debian' to the title of the Ubuntu section since it is applicable there as well.
3. Fixed the heading level for Ubuntu to match the rest of the readme.

## Motivation and Context
This provides Fedora Pihole user with accurate instructions on how to free port 53, which they do not currently have.
[Existing Issue.](https://github.com/pi-hole/docker-pi-hole/issues/1656)

## How Has This Been Tested?
This does not affect any other project code. I have verified that the included commands work as expected on a Fedora 41 server.

## Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
- [X] My code follows the code style of this project.
- [X] My change requires a change to the documentation.
- [X] I have updated the documentation accordingly.

PR #1860, opened 2025-07-06
docs: Add instructions for disabling resolved port 53 for fedora

### LINKED ISSUE #1656: Documentation Error for Fedora Usage
This is a: Documentation Error

## Details

This section is inaccurate on modern Fedora, it might be written with Ubuntu in mind
https://github.com/pi-hole/docker-pi-hole/tree/4cb3c3a16d4d571408f83640487017799402b50b?tab=readme-ov-file#installing-on-ubuntu-or-fedora

There are two issues that stand out to me:
1. As far as I know, modern Fedora does not ship with `netplan`, so `systemd-resolved` on Fedora cannot use it by default. I do not know enough about `netplan` to say if installing it is sufficient to fix this or if that requires configuration.
2. About the line `sudo sed -r -i.orig 's/#?DNSStubListener=yes/DNSStubListener=no/g' /etc/systemd/resolved.conf`; it might be worth mentioning that (as far as I know, at least on some versions), `/etc/systemd/resolved.conf` might not exist by default; a template for it can be copied into `/etc/systemd/resolved.conf` from `/usr/lib/systemd/resolved.conf`. I believe this change was made in Fedora 40? I think a better option would just be:
```bash
$ [ -d /etc/systemd/resolved.conf.d ] || sudo mkdir /etc/systemd/resolved.conf.d
$ echo '[Resolve]' | sudo tee /etc/systemd/resolved.conf.d/10-pihole.conf
$ echo 'DNSStubListener=no' | sudo tee -a /etc/systemd/resolved.conf.d/10-pihole.conf
```

This stack overflow post seems to indicate that
https://unix.stackexchange.com/questions/761092/fedora-systemd-resolved

## Related Issues
- [x] I have searched this repository/Pi-hole forums for existing issues and pull requests that look similar 

This #1365 is about a README overhaul in general, but the description is mostly empty so I am assuming it is not talking about this issue

### Environment data
  * Operating Systems: Fedora Workstation 40, Fedora Cloud 41, Fedora Minimal 41

LizenzFass78851:
## Description
This supplements the instructions for deactivating systemctl resolve port 53 under Fedora, since one of the previous commands only works under Ubuntu (Debian).

## Motivation and Context
- Fixes #1656
- the own continuation of #1660
- to keep the documentation (readme.md) regarding disabling port 53, which is used by systemctl-resolve, up to date.

## How Has This Been Tested?
I created 4 VMs:
- Fedora 33
- Fedora 42
- Ubuntu 18.04
- Ubuntu 24.04

I tried the instructions for disabling systemd-resolve in all 4 VMs and found that, following the previous instructions, this no longer worked on Fedora 42. However, it did work on the other 3 VMs (as also noted in the linked issue).

After updating the documentation (readme.md), I tried the additions on Fedora 42 and port 53 was successfully disabled.

## Types of changes

## Checklist:


README.md
```diff
@@ -186,10 +186,10 @@ Here is a rundown of other arguments for your docker-compose / docker run.
 ### Installing on Ubuntu or Fedora
 
 Modern releases of Ubuntu (17.10+) and Fedora (33+) include [`systemd-resolved`](http://manpages.ubuntu.com/manpages/bionic/man8/systemd-resolved.service.8.html) which is configured by default to implement a caching DNS stub resolver. This will prevent pi-hole from listening on port 53.
-The stub resolver should be disabled with: `sudo sed -r -i.orig 's/#?DNSStubListener=yes/DNSStubListener=no/g' /etc/systemd/resolved.conf`.
+The stub resolver should be disabled with: `sudo sh -c 'mkdir -p /etc/systemd/resolved.conf.d && printf "[Resolve]\nDNSStubListener=no\n" | tee /etc/systemd/resolved.conf.d/no-stub.conf'`.
```
yubiuser, line 189: I like this a lot, as it does what we also do on bare metal installations

https://github.com/pi-hole/pi-hole/blob/3977a312f01687bcc649365297c6a0a3ba402317/automated%20install/basic-install.sh#L1300-L1305

We probably need to reload `systemd-resolved` also to take the new configuration

README.md
```diff
@@ -208,6 +208,13 @@ network:
     version: 2
 ```
 
+For fedora users, you can run the following commands to edit the sysconfig(s) found at `/etc/NetworkManager/system-connections` via nmcli.
+
+Example sysconfig nmcli commands:
```
yubiuser, line 213: I'm not sure if we should have instructions how to set Pi-hole as the DNS for the host at all. While this is a valid configurations, the host's DNS will break if Pi-hole is not running.

yubiuser (commented):
I still think all this configurations is better placed in https://docs.pi-hole.net/docker/

README.md
yubiuser: I see where you'e coming from. This is a more general discussion. (I think I meant to suggest to also move the Ubuntu part to the docs). Maybe @pi-hole/docker-maintainers have an opinion

yubiuser:
Doc PR moving this to the other repo: https://github.com/pi-hole/docs/pull/1266

I thin we can remove the whole section then and only leave a link.

LizenzFass78851:
> I thin we can remove the whole section then and only leave a link.

The pr https://github.com/pi-hole/docker-pi-hole/pull/1863 to remove the section has been created

yubiuser:
https://github.com/pi-hole/docker-pi-hole/pull/1863 has been merged.

PR #1866, opened 2025-07-14
Prevent cosmetic issue where component versions are sometimes labeled as commit hashes

PromoFaux:

Occasionally core and web `master` branches can have additional commits ahead of the latest tagged commit - not often, but often enough to make me kick myself when I tag this repo.

This copies the behaviour of bare metal, where the local repository is hard reset to the latest tag.

---

---

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -60,9 +60,12 @@ COPY crontab.txt /crontab.txt
 # Add PADD to the container, too.
 ADD --chmod=0755 https://raw.githubusercontent.com/${PADD_FORK}/PADD/${PADD_BRANCH}/padd.sh /usr/local/bin/padd
 
-# download a the main repos from github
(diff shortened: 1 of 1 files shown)


yubiuser:
This will increase the image size as we don't need a clone depth 5 in most cases. I think the better approach would be to first check for the branch name and if it's master only clone latest tag; in all other cases only clone depth 1

PromoFaux:
Indeed, by just over 1mb. I can live with that :)

Edit: also, I tried that approach first, but it doesn't work unfortunately. Cloning a specific tag rather than the branch means the repo ends up in a headless state and then the version script does not work.

It's also not possible to change the depth after it has been cloned - which was another approach I looked at.

In the end, I realised i was starting to add a lot of code just to save 1MB

yubiuser:
> Cloning a specific tag rather than the branch means the repo ends up in a headless state and then the version script does not work.

The trick is to checkout a fake 'master' branch

https://github.com/pi-hole/docker-pi-hole/pull/1869

PR #1869, opened 2025-07-16
Only clone depth 1, even when checking out tags

yubiuser:
Builds upon https://github.com/pi-hole/docker-pi-hole/pull/1866.

Instead of cloning master with depth of 5 and checking out the lastest tag 'backwards' we now checkout the tag directly. To fool the version script we need to fake checkout 'master' branch. 

Saves 1 MB.

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -65,15 +65,15 @@ ADD --chmod=0755 https://raw.githubusercontent.com/${PADD_FORK}/PADD/${PADD_BRAN
 # (we need to create a new "master" branch to avoid the "detached HEAD" state for the version check to work correctly)                                      
 
 RUN clone_repo() { \
-        REPO_FORK="$1"; \
-        REPO_NAME="$2"; \
+        FORK="$1"; \
+        REPO="$2"; \
         BRANCH="$3"; \
         DEST="$4"; \
         CLONE_BRANCH="$BRANCH"; \
         if [ "$BRANCH" = "master" ]; then \
-            CLONE_BRANCH=$(curl -s https://api.github.com/repos/${REPO_FORK}/${REPO_NAME}/releases/latest | jq -r .tag_name); \
+            CLONE_BRANCH=$(curl -s https://api.github.com/repos/${FORK}/${REPO}/releases/latest | jq -r .tag_name); \
         fi; \
-        git clone --branch "$CLONE_BRANCH" --single-branch --depth 1 "https://github.com/${REPO_FORK}/${REPO_NAME}.git" "$DEST"; \
+        git clone --branch "$CLONE_BRANCH" --single-branch --depth 1 "https://github.com/${FORK}/${REPO}.git" "$DEST"; \
         cd "$DEST"; \
         if [ "$BRANCH" = "master" ]; then git checkout -b master; fi; \
     }; \


src/Dockerfile
```diff
@@ -61,11 +61,22 @@ COPY crontab.txt /crontab.txt
...
-#if the branch is master we reset to the latest tag as sometimes the master branch contains meta changes that have not been tagged.                                        
-RUN git clone --depth 5 --single-branch --branch ${WEB_BRANCH} https://github.com/${WEB_FORK}/web.git /var/www/html/admin && \    
-    if [ "${WEB_BRANCH}" = "master" ]; then cd /var/www/html/admin && git reset --hard "$(git describe --tags --abbrev=0)"; fi && \
-    git clone --depth 5 --single-branch --branch ${CORE_BRANCH} https://github.com/${CORE_FORK}/pi-hole.git /etc/.pihole && \    
-    if [ "${CORE_BRANCH}" = "master" ]; then cd /etc/.pihole && git reset --hard "$(git describe --tags --abbrev=0)"; fi
+# if the branch is master we clone the latest as sometimes the master branch contains meta changes that have not been tagged
+# (we need to create a new "master" branch to avoid the "detached HEAD" state for the version check to work correctly)                                      
+RUN if [ "${WEB_BRANCH}" = "master" ]; then \
+        LATEST_TAG=$(git ls-remote --tags --sort="v:refname" https://github.com/${WEB_FORK}/web.git | tail -n1 | sed 's/.*\///; s/\^{}//') && \
```
rdwebdesign, line 67: We can avoid tags containing `{}` using [`--refs` option](https://git-scm.com/docs/git-ls-remote#Documentation/git-ls-remote.txt---refs):

```suggestion
        LATEST_TAG=$(git ls-remote --tags --refs --sort="v:refname" https://github.com/${WEB_FORK}/web.git | tail -n1 | sed 's/.*\///') && \
```

rdwebdesign: Or we could simply use Github API to return the tag from the latest release:

```suggestion
        LATEST_TAG=$(curl https://api.github.com/repos/${WEB_FORK}/web/releases/latest | jq -r .tag_name) && \
```

rdwebdesign (changes requested):

PromoFaux:
Actually... or to further reduce duplication:

```dockerfile
RUN clone_repo() { \
        REPO_FORK="$1"; \
        REPO_NAME="$2"; \
        BRANCH="$3"; \
        DEST="$4"; \
        CLONE_BRANCH="$BRANCH"; \
        if [ "$BRANCH" = "master" ]; then \
            CLONE_BRANCH=$(curl -s https://api.github.com/repos/${REPO_FORK}/${REPO_NAME}/releases/latest | jq -r .tag_name); \
        fi; \
        git clone --branch "$CLONE_BRANCH" --single-branch --depth 1 "https://github.com/${REPO_FORK}/${REPO_NAME}.git" "$DEST"; \
        cd "$DEST"; \
        if [ "$BRANCH" = "master" ]; then git checkout -b master; fi; \
    }; \
    clone_repo "${WEB_FORK}" "web" "${WEB_BRANCH}" "/var/www/html/admin"; \
    clone_repo "${CORE_FORK}" "pi-hole" "${CORE_BRANCH}" "/etc/.pihole"
```

rdwebdesign:
I like the last suggestion using a function.

PR #1880, opened 2025-08-06
Add pihole-FTL bash completion

yubiuser:
Adds the FTL bash completion to the image. Depends on https://github.com/pi-hole/pi-hole/pull/6376

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -90,7 +90,8 @@ RUN cd /etc/.pihole && \
     install -Dm755 -d /var/log/pihole && \
     install -Dm755 -d /var/lib/logrotate && \
     install -Dm755 -t /usr/local/bin pihole && \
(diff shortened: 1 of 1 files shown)


yubiuser:
Un-draft once the `core` PR has been merged. Otherwise test will always fail.

yubiuser:
Core and FTL branch have been merged into development.

rdwebdesign:
Are you sure this is needed?

I think the Core PR will already install the files.

yubiuser:
If we want the feature on docker, we need this PR. On bare metall the completion file is installed with the installer, but on docker we don't run the installer and therefore need to copy the file manually.
We do the same for 'pihole' bash completion already

PromoFaux (approved):
This should also fix failing builds

PR #1898, opened 2025-09-24
Use FTL's build-in wait-for command for checking the logs

yubiuser:
Use `pihole-FTL wait-for` function to wait for a specific log line to appear.

https://github.com/pi-hole/FTL/pull/2631

### DIFF
--- a/src/start.sh
+++ b/src/start.sh
@@ -75,10 +75,9 @@ start() {
         sleep 0.5
     done
 
-    #  Wait until the FTL log contains the "FTL started" message before continuing
-    while ! grep -q '########## FTL started' /var/log/pihole/FTL.log; do
-        sleep 0.5
-    done
+    #  Wait until the FTL log contains the "FTL started" message before continuing, timeout after 10 seconds
+    logsize_before=$(stat -c%s /var/log/pihole/FTL.log)
+    /usr/bin/pihole-FTL wait-for '########## FTL started' /var/log/pihole/FTL.log 10 "${logsize_before}"
 
     pihole updatechecker
     local versionsOutput
--- a/src/start.sh
+++ b/src/start.sh
@@ -77,7 +77,7 @@ start() {
 
     #  Wait until the FTL log contains the "FTL started" message before continuing, timeout after 10 seconds
     logsize_before=$(stat -c%s /var/log/pihole/FTL.log)
-    /usr/bin/pihole-FTL wait-for '########## FTL started' /var/log/pihole/FTL.log 10 "${logsize_before}"
+    pihole-FTL wait-for '########## FTL started' /var/log/pihole/FTL.log 10 "${logsize_before}" > /dev/null
 
     pihole updatechecker
     local versionsOutput
--- a/test/tests/test_general.py
+++ b/test/tests/test_general.py
@@ -50,17 +50,19 @@ def test_pihole_ftl_architecture(docker):
 
 
 # Wait 5 seconds for startup, then kill the start.sh script
-# Finally, grep the FTL log to see if it has been shut down cleanly
+# Finally, check the FTL log to see if it has been shut down cleanly
 def test_pihole_ftl_clean_shutdown(docker):
     func = docker.run(
         """
         sleep 5
         killall --signal 15 start.sh
-        sleep 5
-        grep 'terminated' /var/log/pihole/FTL.log
+        pihole-FTL wait-for '########## FTL terminated after' /var/log/pihole/FTL.log 10
     """
     )
-    assert "INFO: ########## FTL terminated after" in func.stdout
+    assert (
+        'Found string "########## FTL terminated after" in file /var/log/pihole/FTL.log'
+        in func.stdout
+    )
     assert "(code 0)" in func.stdout
 
 


yubiuser:
Tests are failing because the line ` ########## FTL terminated` is not printed anymore. This also happens on bare metal `FTL's` `development` branch.
@DL6ER

DL6ER:
Yes, this is an interesting bug. We `free` the location where `files.log.ftl` is stored and thereafter, no new lines are printed to the file. We obviously need an exception here. This is a regression of my recent fixing-all-memory-leaks PR. Seem we want to actually leak at least this single bit of memory.

DL6ER:
See https://github.com/pi-hole/FTL/pull/2634

yubiuser:
I tried to use `pihole-FTL wait-for` also in the tests at 

https://github.com/pi-hole/docker-pi-hole/blob/5ff925b66f9ec65f1b24ad28c4afdb7b0927f0d6/test/tests/test_general.py#L59-L60

but it did not work. The `pihole-FTL wait-for` command did not produce any output (which it should even if it did not find the string), I guess it's because `start.sh` was already killed. It works with the `grep` version we use so far so I decided not to invest too much time.

src/start.sh
```diff
@@ -75,10 +75,9 @@ start() {
         sleep 0.5
     done
 
-    #  Wait until the FTL log contains the "FTL started" message before continuing
-    while ! grep -q '########## FTL started' /var/log/pihole/FTL.log; do
-        sleep 0.5
-    done
+    #  Wait until the FTL log contains the "FTL started" message before continuing, timeout after 10 seconds
+    logsize_before=$(stat -c%s /var/log/pihole/FTL.log)
```
PromoFaux, line 79: Do we need to get this?

Reading the comments in the FTL PR, we can just pass in 0 and it will start reading from the end of the file. I feel like not much time will pass between lines 79 and 80 it worth populating the variable?

src/start.sh
```diff
@@ -75,10 +75,9 @@ start() {
...
+    pihole-FTL wait-for '########## FTL started' /var/log/pihole/FTL.log 10 "${logsize_before}" > /dev/null
```
PromoFaux, line 80: _Something_, and I don't know what, is giving me pause on the 10 second timeout here. 

While it _shouldn't_ take anywhere near 10 seconds for FTL to start up, if (for whatever reason) it _doesn't_ start up then the script will continue on as though it _has_. Currently we go around forever and ever and ever and ever - so it's an improvement for sure - but maybe this logic can be shored up a little

PromoFaux:
> but it did not work. ...[snip]... I guess it's because `start.sh` was already killed.

This is odd to me, too. `start.sh` shouldn't come into it here? It's not so important, but it is perplexing!!

I'm also wondering if a "better" test for clean shut down is actually volume mounting the logs directory and then checking the contents of the log file _after_ the container has stopped. But that's for another time.

src/start.sh
yubiuser: Probably we don't need this.

src/start.sh
yubiuser: We could exit if the return code is != 0 indicating that is has not been found

yubiuser:
> I'm also wondering if a "better" test for clean shut down is actually volume mounting the logs directory and then checking the contents of the log file after the container has stopped. But that's for another time.

I think you're onto something here. We should not try to postpone shutdown just to read logs from inside the container.

src/start.sh
```diff
@@ -75,10 +75,13 @@ start() {
...
 
-    #  Wait until the FTL log contains the "FTL started" message before continuing
-    while ! grep -q '########## FTL started' /var/log/pihole/FTL.log; do
-        sleep 0.5
-    done
+    # Wait until the FTL log contains the "FTL started" message before continuing, timeout after 10 seconds
+    # exit if we do not find it
+    pihole-FTL wait-for '########## FTL started' /var/log/pihole/FTL.log 10 "0 > /dev/null"
+    if [ $? -ne 0 ]; then
+        echo "  [✗] FTL did not start"
```
PromoFaux, line 82: I'm just being pernickety at this point, but can we tack on something like ", stopping container" here. Not 100% on the wording, but i'd like just a little more verbosity in the sentence

```suggestion
        echo "  [✗] FTL did not start - stopping container"
```

yubiuser:
>  But that's for another time.

https://github.com/pi-hole/docker-pi-hole/pull/1899

src/start.sh
```diff
@@ -75,10 +75,13 @@ start() {
         sleep 0.5
     done
 
-    #  Wait until the FTL log contains the "FTL started" message before continuing
-    while ! grep -q '########## FTL started' /var/log/pihole/FTL.log; do
-        sleep 0.5
-    done
+    # Wait until the FTL log contains the "FTL started" message before continuing, timeout after 10 seconds
+    # exit if we do not find it
+    pihole-FTL wait-for '########## FTL started' /var/log/pihole/FTL.log 10 "0 > /dev/null"
```
PromoFaux, line 80: On the PR for the shutdown test - you comment that we check for up to 60 secodns to give slow RISCV devices a chance to start up. Perhaps we should wait 60 seconds here.

yubiuser: The difference is: Within the `start.sh, ` pihole-FTL wait-for` will only wait for the log line to appear, which is triggered by the ` capsh --user="${DNSMASQ_USER}" --keep=1 -- -c "/usr/bin/pihole-FTL $FTL_CMD >/dev/null" &` line just above. Not much time should be in between. 

The other PR watches the container from the outside before it kills the container. There are other (maintenance) tasks before FTL will even be invoked, e.g. running gravity. Therefore I gave it more time.

PromoFaux (approved):

casperklein:
FTL's build-in wait-for command uses inotify, which doesn't support network file systems.

https://man7.org/linux/man-pages/man7/inotify.7%40%40man-pages.html

>Limitations and caveats
>      The inotify API provides no information about the user or process
>      that triggered the inotify event.  In particular, there is no easy
>      way for a process that is monitoring events via inotify to
>      distinguish events that it triggers itself from those that are
>      triggered by other processes.
>      Inotify reports only events that a user-space program triggers
>      through the filesystem API.  **As a result, it does not catch remote
>      events that occur on network filesystems.  (Applications must fall
>      back to polling the filesystem to catch such events.)**

This could introduce potential issues with network mounted file systems like NFS/CIFS/SMB.

I found this alternative:

```bash
( tail -f -n0 /var/log/pihole/FTL.log & ) | grep -q "########## FTL started"
```

See also https://superuser.com/a/900134

PromoFaux:
Just did a simple test from Docker Desktop on windows - though can't guarantee I've done this the right way, it _appears_ to work fine?

image

casperklein:
Thanks for testing. I did a bit more research. I think this works, because the changes are made locally by FTL. So the kernel knows about this and inotify works.
It does not work, if the log file is altered by another client or on the server directly where the mount is on.

But that shouldn't be an issue here. Sorry for the noise.

In another project we run into issues, when we watched a certificate file for changes on a NFS mount with inotify. The certificate was managed and renewed externally. Inotify did not work in that case. That triggered me when I saw this PR.

yubiuser:
@casperklein Thanks for keeping an eye on this. Network share have been causing trouble before already.

PromoFaux:
No need to apologise for noise!

It's an angle I never thought to test - and I learned something about creating CIFS volumes in docker :)

I tried with an NFS volume too,  but there were also all sorts of other permission errors there, with or without this change set. Either I have no idea what I'm doing with NFS, or it is a tricky beast!

Interestingly, I couldn't use a UNC path or a mapped network drive directly in the compose file.

PR #1899, opened 2025-09-27
Improve/check shutdown

yubiuser:
This PR improves the `test_pihole_ftl_clean_shutdown` function. Instead of shutting down the container from within and delay shutdown to `grep` for log lines, this PR shuts down the container from outside and uses `docker log` to search for the relevant FTL log line.
The second commit removes the `$PYTEST` variable: it was only used to delay the container shutdown during the tests and to skip creation of the default adlist. We can surely debate if we want to keep the variable anyways.

### DIFF
--- a/test/tests/test_general.py
+++ b/test/tests/test_general.py
@@ -49,7 +49,7 @@ def test_pihole_ftl_architecture(docker):
     assert platform in func.stdout
 
 
-# Wait 5 seconds for startup, then stop the container gracefully
+# Wait for FTL to start up, then stop the container gracefully
 # Finally, check the container logs to see if FTL was shut down cleanly
 def test_pihole_ftl_clean_shutdown(docker):
     import subprocess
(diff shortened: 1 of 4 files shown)


test/tests/test_general.py
```diff
@@ -58,8 +58,23 @@ def test_pihole_ftl_clean_shutdown(docker):
     # Get the container ID from the docker fixture
     container_id = docker.backend.name
 
-    # Wait for startup
-    time.sleep(5)
+    # Wait for FTL to fully start up by checking logs
+    max_wait_time = 60  # Maximum wait time in seconds
```
PromoFaux, line 62: Thought (I might be wrong, I'm hungover and on a train):

In the other PR, we wait for 10 seconds for FTL to start - and if it does not, we stop the container. 

As the running code already has a built-in timeout, do we need to do this here too (and for longer...)?

Also maybe the test should be renamed to something like "test_pihole_ftl_starts_and_shuts_down_cleanly"

PromoFaux (dismissed):

PromoFaux:
Ah, needs a little rebase

yubiuser:
Rebased and forced push.

PromoFaux (approved):

PR #1905, opened 2025-10-05
fix: publish docker image by pushing on development branch

LizenzFass78851:
## Description
This adds the variable `components branch` so that if `github.ref_name` is recognized as `development`, it also pushing the docker image to the `development` tag

## Motivation and Context
In one of the last workflow runs that was triggered by pushing on the `development` branch, an error occurred that was triggered when the Docker image was built and not the `development` branch of FTL, Web and Core was loaded, but rather the respective `master` branch.
https://github.com/pi-hole/docker-pi-hole/actions/runs/18255584741/job/51976380837#step:10:399

## How Has This Been Tested?
I tested the workflow with deviations in my fork with the changes

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1906, opened 2025-10-06
Drop the development tag, reduce dupliation in build yaml

PromoFaux:

This PR does two things:
 - Drops the `development` tag. After thinking about it a lot, the `development` tag is next to pointless. And does not build when we end up in situations where the core/web/FTL code has changed in structure between their own master and dev branches. (See https://github.com/pi-hole/docker-pi-hole/pull/1905)
 - Reduces reuse of exact steps the build YAML by utilising YAML anchors, making it easier to maintain, and to reduce the possibility of error when updating two otherwise identical sections.

---

---

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -40,17 +40,20 @@ jobs:
           platform=${{ matrix.platform }}
           echo "PLATFORM_PAIR=${platform//\//-}" >> $GITHUB_ENV
 
-      - name: Checkout Repo
+      - &checkout-repo
+        name: Checkout Repo
         if: github.event_name != 'schedule'
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 #v5.0.0
 
-      - name: Checkout dev branch if scheduled
+      - &checkout-dev
+        name: Checkout dev branch if scheduled
         if: github.event_name == 'schedule'
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 #v5.0.0
         with:
           ref: development
 
-      - name: Docker meta
+      - &docker-meta
+        name: Docker meta
         id: meta
         uses: docker/metadata-action@c1e51972afc2121e065aed6d45c65596fe445f3f #v5.8.0
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-publish.yml
```diff
@@ -40,17 +40,20 @@ jobs:
           platform=${{ matrix.platform }}
           echo "PLATFORM_PAIR=${platform//\//-}" >> $GITHUB_ENV
 
-      - name: Checkout Repo
+      - &checkout-repo
+        name: Checkout Repo
         if: github.event_name != 'schedule'
```
yubiuser, line 45: This should happen only on release.

.github/workflows/build-and-publish.yml
```diff
@@ -40,17 +40,20 @@ jobs:
...
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 #v5.0.0
 
-      - name: Checkout dev branch if scheduled
+      - &checkout-dev
+        name: Checkout dev branch if scheduled
         if: github.event_name == 'schedule'
```
yubiuser, line 50: This should be triggered when event is **not** release

yubiuser (changes requested):

PromoFaux:
We should now see the nightly get updated once I merge this...

yubiuser:
Worked fine. Do we need to un-tag `development` once this has reached master?

PromoFaux:
It can stay there and just get older and older. I was toying with the idea of doing one last update to it to print something like the following on startup. But,.. probably not worth it.

```

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
!! THIS TAG IS DEPRECATED - PLEASE USE NIGHTLY !!
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
```

yubiuser:
I think a note to the readme would be sufficient...

rdwebdesign:
I don't think we should update an image just to say the tag is deprecated and it won't be updated anymore.

> I think a note to the readme would be sufficient

I agree we need a note on the README, but maybe we could retire the tag to make sure future users won't try to start containers using this tag.

Issue #1918, opened 2025-10-26
FTL fails to start after upgrading docker instance from 2025.08.0 (6.1.4) to 2025.10.0 (6.2.0 or 6.2.1 - assuming 6.2.1 since there is no 2025.10.1)


jsnfwlr:
### Versions
- Core: v6.1.4 -> v6.2.1
- Web: v6.2.1 -> v6.3
- FTL : v6.2.3 -> v6.3

### Platform
- OS and version:  Host is running Ubuntu 24.04.3 LTS
- Platform:  Docker

### Expected behavior

Should start, or should provide a meaningful error telling me what when wrong.

### Actual behavior / bug

FTL fails, without meaningful error - verbose container logs show the folllowing:

```
+ sh /opt/pihole/pihole-FTL-prestart.sh
+ echo '  [i] Starting pihole-FTL (no-daemon) as pihole'
+ echo ''
  [i] Starting pihole-FTL (no-daemon) as pihole

+ CAPSH_PID=53
+ '[' '!' -f /var/log/pihole/FTL.log ']'
+ + capsh pihole-FTL --user=pihole wait-for --keep=1 -- '########## FTL started' -c /var/log/pihole/FTL.log 10 0
'/usr/bin/pihole-FTL no-daemon >/dev/null'
+ '[' 1 -ne 0 ']'
+ echo '  [✗] FTL did not start - stopping container'
+ exit 1
```
/var/log/pihole/FTL.log is empty

### Steps to reproduce

Change image address in docker compose from ` 2025.08.0` to ` 2025.10.0` and run `docker compose up` and it fails

## Debug Token
Container won't start so I can't get to the CLI to generate the token

eacunha:
I think I have the "samey" issue (sorry, I am kind of noob about this). I am running Pi-Hole in my raspberry pi5 and using Watchtower to get it updated all the time. Today/yesterday it stopped working. I look at the changelogs of this version and I see a lot of FTL-related updates, so I am am now wondering how to update it. Below is the log from the container:

[i] pihole-FTL pre-start checks
  [i] Setting capabilities on pihole-FTL where possible
  [i] Applying the following caps to pihole-FTL:
        * CAP_CHOWN
        * CAP_NET_BIND_SERVICE
        * CAP_NET_RAW
        * CAP_NET_ADMIN
  [i] Starting pihole-FTL (no-daemon) as pihole
  [✗] FTL did not start - stopping container
[...]

PR #1919, opened 2025-10-26
Increase default FTL Startup wait time to 30 seconds

PromoFaux:

2025.10.0 introduced a change whereby we exit the container if FTL has not started up within 10 seconds. It seems that this may be a little too optimistic. See https://github.com/pi-hole/docker-pi-hole/issues/1918

This PR increases the default timeout to 30 seconds, which should surely be enough... but if it isn't, we also introduce an environment variable named FTL_START_TIMEOUT for the user to set the value to their liking, depending on their system behaviour.

An example of the container exit message shown below, with FTL_START_TIMEOUT set to 0 seconds to demonstrate behaviour

image

---

---

### DIFF
--- a/src/start.sh
+++ b/src/start.sh
@@ -75,9 +75,12 @@ start() {
         sleep 0.5
     done
 
-    # Wait until the FTL log contains the "FTL started" message before continuing, timeout after 10 seconds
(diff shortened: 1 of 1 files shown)


yubiuser:
This variable should be added to README

PR #1920, opened 2025-10-26
More helpful error logging on startup

PromoFaux:
Draft...

### DIFF
--- a/src/start.sh
+++ b/src/start.sh
@@ -80,10 +80,11 @@ start() {
 
     # Wait until the FTL log contains the "FTL started" message before continuing
     # exit if we do not find it
-    pihole-FTL wait-for '########## FTL started' /var/log/pihole/FTL.log "${FTL_START_TIMEOUT}" 0 > /dev/null
+    local wait_output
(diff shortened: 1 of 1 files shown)


PromoFaux:
@ICreatedANewAccountJustForThis - are you able to checkout and build this branch locally and report the error message that is displayed. If it is anyhting other than "Did not find string "########## FTL started" in file /var/log/pihole/FTL.log within x seconds" then we have another issue :)

casperklein:
You could also add

`grep '########## FTL started' /var/log/pihole/FTL.log`

below

`echo "  [✗] ${wait_output}"`

to see, if FTL is started, but not recognized by `pihole-FTL wait-for` (inotify may not working as expected).

Edit: just noticed, [you already proofed that](https://github.com/pi-hole/docker-pi-hole/issues/1918#issuecomment-3448489763)

PromoFaux:
We're just going to [revert the change](https://github.com/pi-hole/docker-pi-hole/pull/1922) for the time being, it seems that `pihole-FTL wait-for` is not working as expected.

PR #1934, opened 2025-11-03
Redo of using pihole-FTL wait-for the right way

yubiuser:
This is a redo of https://github.com/pi-hole/docker-pi-hole/pull/1898 which caused issues like https://github.com/pi-hole/docker-pi-hole/issues/1918.

As an interim solution, we reverted the original PR - but I could not resist to try it again. Reviewing https://github.com/pi-hole/docker-pi-hole/pull/1932 and reading the commit message of `[this commit](https://github.com/pi-hole/docker-pi-hole/pull/1932/commits/9e8a13225013940878a6a255306f9c1cdbfe2325)`, it became quite clear what I did wrong: 

The `wait-for` function uses a parameter which indicates a starting point to watch the file for changes. I misinterpreted the  `0` as 'start from beginning of the file' where it really is 'end of the file'. See [here](https://github.com/pi-hole/FTL/blob/bc5e8b22a3e55db256a634743ebf4855e5e149e6/src/config/inotify.c#L171). 
The problem is: when FTL starts really fast, the log will contain the startup message, before we invoke the `wait-for` function which will time out, as it watches the log after the line has already been printed. 
The solution is pretty simple: get the file size before starting FTL and inspect the file from this point.

Added @casperklein as Co-author as they gave the right clue.

### DIFF
--- a/src/start.sh
+++ b/src/start.sh
@@ -58,6 +58,9 @@ start() {
     fix_capabilities
     sh /opt/pihole/pihole-FTL-prestart.sh
 
+    # Get the file size of the FTL log before starting FTL
+    logsize_before=$(stat -c%s /var/log/pihole/FTL.log)
+
     echo "  [i] Starting pihole-FTL ($FTL_CMD) as ${DNSMASQ_USER}"
     echo ""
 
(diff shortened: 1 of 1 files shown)


yubiuser:
RISCV test failing due to external issue

```
#12 2.870 (19/74) Installing bind-libs (9.20.15-r0)
#12 3.146 (20/74) Installing bind-tools (9.20.15-r0)
#12 3.175 ERROR: bind-tools-9.20.15-r0: BAD signature
#12 3.175 (21/74) Installing jansson (2.14.1-r0)
#12 3.508 (22/74) Installing zstd-libs (1.5.7-r0)
```

I think we have to re-run this test once they fixed it on their end

casperklein:
I was thinking about closing my PR in favor of this one.

My PR contains also some other improvements:

- using `startFrom` variable instead of `logsize_before` because the variable is later used for a similar purpose. No need to introduce a new variable.
- Using $FTLlogFile instead of the hard-coded log path
- Added `--` to grep/tail commands as best-practice to prevent issues with (weird) file names like `--FTL.log`.

I can offer to replace the `tail/grep` part with your `pihole-FTL wait-for` approach in my PR.

yubiuser:
Sure, go ahead. Just note that `pihole-FTL wait-for` expects a file size, not a log line number as starting point.

PromoFaux:
Which one am I reviewing? 😅

yubiuser:
If it works, go for @casperklein's.

PromoFaux:
Closing both this and #1932 in favour of https://github.com/pi-hole/docker-pi-hole/pull/1940 (everyone is credited!)

Issue #1937, opened 2025-11-05
`riscv64` image no longer building via github actions


PromoFaux:
Build logs suggest and issue with `bind-tools`:

```
#11 2.874 (20/74) Installing bind-tools (9.20.15-r0)
#11 2.906 ERROR: bind-tools-9.20.15-r0: BAD signature
```

But I don't have time to look into it currently. It's already way past my bedtime.

yubiuser:
I had the same here: https://github.com/pi-hole/docker-pi-hole/pull/1934, but re-started RISCV yesterday and it went through. Strange.

PromoFaux:
The actual build and merge was failing consistently yesterday at installing `bind-tools`, despite the tests build succeeding. 

Before last night the nightly had not been updated in 4 days, either - so clearly something has gone wrong somewhere!

PromoFaux:
Though I just kicked this previously failing buiild and it worked this time.

https://github.com/pi-hole/docker-pi-hole/actions/runs/19087309414/job/54585877486

yubiuser:
> The actual build and merge was failing consistently yesterday at installing bind-tools, despite the tests build succeeding.

This should not happen. It should always pass or always fail....

rdwebdesign:
This is not consistently happening. Sometimes it works, sometimes it fails.

Not sure when it started.

Yesterday I also tried to [re-run the job](https://github.com/pi-hole/docker-pi-hole/actions/runs/19081700346), but it failed again.

Initially, I thought this was caused by some external failure, but I'm not sure what is happening.

PromoFaux:
It's clearly transient. Maybe we just have to retry it a few times when it happens. 

https://github.com/pi-hole/docker-pi-hole/actions/runs/19115751605?pr=1932

yubiuser:
Issue should be fixed with https://github.com/pi-hole/docker-pi-hole/releases/tag/2026.02.0

PR #1939, opened 2025-11-05
Add RISCV back

yubiuser:
For tracking.
Fixes https://github.com/pi-hole/docker-pi-hole/issues/1937


PromoFaux:
It must have been having a funny 5 minutes. It is now building fine.

PR #1987, opened 2026-02-14
Set fixed buildx version to mitigate issues with buildx version 0.31.1

yubiuser:
Since 6th of feb our publish workflow failed without us having changed anything. Deep inspection revealed, that the buildx version installed by `docker/setup-buildx-action` increased from `v0.31.0` to `v0.31.1`. The new version made changes to the command we use `buildx imagetools create`. (See https://github.com/docker/buildx/releases/tag/v0.31.1).

This PR set the installed version to a fixed version  `v0.31.0` to hopefully mitigate the issue. This change need to be reverted in the future when things have been sorted at `buildx` side.

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -95,6 +95,10 @@ jobs:
       - &setup-buildx
         name: Set up Docker Buildx
         uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f #v3.12.0
+        with:
+          # Buildx version 0.31.1 broke our publish workflow, this need to be revised when 0.32.0 is released
+          # https://github.com/docker/buildx/releases/tag/v0.31.1
+          version: v0.31.0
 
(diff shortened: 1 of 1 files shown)


PromoFaux:
I've clearly been smoking crack. 



<strike>It does not appear to be that which is the issue. Cherry-picked this commit to the `experimental` branch and we observe the same effects:

https://github.com/pi-hole/docker-pi-hole/actions/runs/22039441961/job/63677835687

However, like with the `nightly` tag that is currently "failing"  the images are actually pushed to both hub and GHCR:

https://hub.docker.com/layers/pihole/pihole/experimental/images/sha256-0cbdf29ec15b4b91f126883a37283e077dbb4c6ad29aac7cb0c80741632f946a

https://github.com/pi-hole/docker-pi-hole/pkgs/container/pihole/687630917?tag=experimental</strike>

yubiuser:
Created an issue upstreams: https://github.com/docker/buildx/issues/3708

PR #2032, opened 2026-04-28
Add `gawk` to dependencies

yubiuser:
Just to be save for https://github.com/pi-hole/pi-hole/pull/6621

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -27,6 +27,7 @@ RUN apk add --no-cache \
     binutils \
     coreutils \
     curl \
+    gawk \
     git \
     # Install grep to avoid issues in pihole -w/b with the default busybox grep
     grep \


rdwebdesign:
@yubiuser 

I don't think this is needed.

`awk` was always used in Docker images, a long time before _PR6621_.

We never had issues with the default installed version (busybox).

yubiuser:
I see your point and I'm a bit undecided here. Sure, we never had issues so far, but I'd like to be explicit with our dependencies and be able to use the same options/functions in all scripts across all distros.

rdwebdesign:
I just don't want to increase the image size too much.

I think adding `gawk` will increase the image size in aprox. 2MB. If we decide to add all dependencies (to avoid using busybox versions), the image size will be a log bigger.

rdwebdesign:
> but I'd like to be explicit with our dependencies and be able to use the same options/functions in all scripts across all distros.

I agree this is probably needed for all supported distros (bare metal installs), since we don't control them, but the docker image packages are controlled by us.
We don't need unnecessary dependencies just to match the packages installed on other distros.

yubiuser:
> I agree this is probably needed for all supported distros (bare metal installs), since we don't control them, but the docker image packages are controlled by us.

Ok, this convinced me.

PR #2035, opened 2026-05-16
Make crond and cron job output visible in "docker log"

casperklein:
## Description

Currently `crond` runs silently in the background. There is no indication, when jobs are running or if the jobs produce any output (in case of failures).

This PR solves this by running `crond` with increased log level (6). It also prefixes each line with the current date/time/timezone to align with FTL's output.

The result in "docker log" looks like this:

```
2026-05-16 13:37:38 CEST crond: crond (busybox 1.37.0) started, log level 6
[..]
2026-05-16 13:37:38.899 CEST [52/T59] INFO: Web server ports:
2026-05-16 13:37:38.899 CEST [52/T59] INFO:   - 0.0.0.0:80 (HTTP, IPv4, optional, OK)
2026-05-16 13:37:38.899 CEST [52/T59] INFO:   - 0.0.0.0:443 (HTTPS, IPv4, optional, OK)
2026-05-16 13:37:38.899 CEST [52/T59] INFO:   - [::]:80 (HTTP, IPv6, optional, OK)
2026-05-16 13:37:38.899 CEST [52/T59] INFO:   - [::]:443 (HTTPS, IPv6, optional, OK)
2026-05-16 13:37:38.900 CEST [52/T55] INFO: Compiled 0 allow and 0 deny regex for 0 client in 0.0 msec
2026-05-16 13:37:53.067 CEST [52/T54] INFO: Received 8/8 valid NTP replies from pool.ntp.org
2026-05-16 13:37:53.068 CEST [52/T54] INFO: NTP time offset: 1.463992e+00 ms (excluded 1 outliers)
2026-05-16 13:37:53.068 CEST [52/T54] INFO: NTP round-trip delay: 1.893868e+01 ms (excluded 1 outliers)
2026-05-16 13:37:53.068 CEST [52/T148] INFO: NTP server listening on 0.0.0.0:123 (IPv4)
2026-05-16 13:37:53.068 CEST [52/T149] INFO: NTP server listening on :::123 (IPv6)
[..]
2026-05-16 13:38:00 CEST crond: USER root pid 150 cmd PATH="$PATH:/usr/sbin:/usr/local/bin/" pihole updatechecker
```

## How Has This Been Tested?

Yes, with a local build to produce the output above.

## Types of changes

## Checklist:

### DIFF
--- a/src/prefix-time.sh
+++ b/src/prefix-time.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+# Prefix each line from STDIN with the current date/time/timezone and write to PID 1 STDOUT (docker log)
+while IFS= read -r line; do
+    printf '%s %s\n' "$(date '+%F %T.%3N %Z')" "$line" >>/proc/1/fd/1
+done
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -98,6 +98,7 @@ RUN cd /etc/.pihole && \
     echo "${PIHOLE_DOCKER_TAG}" > /pihole.docker.tag
 
 COPY --chmod=0755 bash_functions.sh /usr/bin/bash_functions.sh
+COPY --chmod=0755 prefix-time.sh /usr/bin/prefix-time.sh
 COPY --chmod=0755 start.sh /usr/bin/start.sh
 
 EXPOSE 53 53/udp
--- a/src/bash_functions.sh
+++ b/src/bash_functions.sh
@@ -89,7 +89,9 @@ start_cron() {
         echo "  [!] Failed to install crontab - scheduled tasks (gravity, update checker) will not run"
     fi
 
-    /usr/sbin/crond
+    # Run crond in foreground, prefix each line from STDIN/STDOUT with the current date/time/timezone and
+    # write to PID 1 STDOUT (docker log)
+    { /usr/sbin/crond -f -d 6 |& prefix-time.sh; } &
     echo ""
 }
 


yubiuser (dismissed):
Nice change. The first one is off a bit, but I guess that's what it is

```
pihole  |   [i] Assigning password defined by Environment Variable
pihole  |   [i] Starting crond for scheduled scripts. Randomizing times for gravity and update checker
pihole  | 
pihole  |   [i] Ensuring logrotate script exists in /etc/pihole
pihole  | 2026-05-16 12:07:37 UTC crond: crond (busybox 1.37.0) started, log level 6
pihole  | 
pihole  |   [i] Gravity migration checks

```

src/prefix-time.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+# Prefix each line from STDIN with the current date/time/timezone and write to PID 1 STDOUT (docker log)
+while IFS= read -r line; do
+    printf '%(%F %T %Z)T %s\n' -1 "$line" >>/proc/1/fd/1
```
PromoFaux:
```suggestion
    printf '   %(%F %T %Z)T %s\n' -1 "$line" >>/proc/1/fd/1
```

Would this sort the offset issue @yubiuser pointed out?

casperklein: Yes and no. If spaces are added like in your suggestion, then the first line would be fine, but all following lines would be off:

```
2026-05-16 13:37:53.068 CEST [52/T148] INFO: NTP server listening on 0.0.0.0:123 (IPv4)
2026-05-16 13:37:53.068 CEST [52/T149] INFO: NTP server listening on :::123 (IPv6)
   2026-05-16 13:38:00 CEST crond: USER root pid 150 cmd PATH="$PATH:/usr/sbin:/usr/local/bin/" pihole updatechecker
```

The only solution for this would be, that `prefix-time.sh` handles the first input line differently than the others.

The reason why I didn't implement it that way from the beginning was, that I wanted a small and simple solution. And that "prefix-time.sh" does exactly what it's name suggests. Maybe the script will be reused in the future for other things?

> The only solution for this would be, that `prefix-time.sh` handles the first input line differently than the others.

I can do that. Just let me know what you think.

PromoFaux: I would expect the first line and subsequent lines to match regardless, that's really weird!

If we could make it match up, that would be great - but only if it doesn't overcomplicate things.

One more small nitpick - I guess `printf` does not support sub-second precision, but maybe something like:

```bash 
printf '%s %s\n' "$(date '+%F %T.%3N %Z')" "$line"
```

The datetime would then be the same as FTL

rdwebdesign: > I would expect the first line and subsequent lines to match regardless, that's really weird!

The misalignment here has a different cause.

If we use your suggestion, all lines printed by `prefix-time.sh` would be indented (as you expected), but there are lines printed by other functions.

<details>
<summary>Details</summary>

https://github.com/pi-hole/docker-pi-hole/blob/dc86ea3dadbcc366f1a9c40f2bfbeb8184aeb7ef/src/bash_functions.sh#L83

https://github.com/pi-hole/docker-pi-hole/blob/dc86ea3dadbcc366f1a9c40f2bfbeb8184aeb7ef/src/bash_functions.sh#L105-L116

These lines start with 2 spaces, hardcoded in the `echo` command, but even in this case, some functions have a multi-line output, like `install_additional_packages()`, causing misalignment

image

Also, there are FTL.log lines printed by `tail` command.


image
</details>

casperklein: > One more small nitpick - I guess printf does not support sub-second precision, but maybe something like:

I tried that before in a vanilla alpine container and it didn't work. I tested again and in docker-pi-hole however it works. I found out, that the `coreutils` package adds the support.

I've applied your suggestion 👍

yubiuser (approved):
```
pihole  | 2026-06-02 11:48:33.939 UTC [55/T56] INFO: NTP time offset: -6.914139e-02 ms (excluded 0 outliers)
pihole  | 2026-06-02 11:48:33.939 UTC [55/T56] INFO: NTP round-trip delay: 2.801740e+01 ms (excluded 0 outliers)
pihole  | 2026-06-02 11:48:33.940 UTC [55/T150] INFO: NTP server listening on 0.0.0.0:123 (IPv4)
pihole  | 2026-06-02 11:48:33.940 UTC [55/T151] INFO: NTP server listening on :::123 (IPv6)
pihole  | 2026-06-02 11:49:00.622 UTC crond: USER root pid 161 cmd PATH="$PATH:/usr/sbin:/usr/local/bin/" pihole updatechecker

```

I think it's fine to not handle the first `crond` line separately, no end user will probably notice and complain.

yubiuser:
Once we mangle it into `json`, output format becomes irrelevant ;-)

```
pihole  | {"timestamp":"2026-06-02 19:00:16.237 UTC","log_level":"INFO","service":"pihole-docker","message":"Ensuring logrotate script exists in /etc/pihole"}
pihole  | {"timestamp":"2026-06-02 19:00:16.237 UTC","log_level":"INFO","service":"crond","message":"crond: crond (busybox 1.37.0) started, log level 6"}
pihole  | {"timestamp":"2026-06-02 19:00:16.247 UTC","log_level":"INFO","service":"pihole-docker","message":"Gravity migration checks"}
```

PR #2038, opened 2026-05-24
feat: optimize Docker build caching

lightswitch05:
## Description
- Use a BuildKit cache mount (`--mount=type=cache,target=/var/cache/apk`) for `apk add` so downloaded packages are reused across builds instead of re-fetched every time
- Add `--link` flag to `ADD` instructions (macvendor.db, padd.sh) so Docker can reorder and restore those layers independently, improving cache reuse
- Enable `cache-mode: max` in the CI workflow to maximize the number of layers saved to the cache
- Reorder Dockerfile layers to place volatile steps (nightly debug tools, crontab copy, docker tag file) after the stable install layers, so a nightly-only change doesn't bust the cache for normal builds

## Motivation and Context
Allowing individual layers to be cached between builds means smaller quicker image pulls between versions. For example, if there is no PADD change, then that layer can get reused between releases.

## How Has This Been Tested?

For testing, I ran the pipeline twice. The [first run](https://github.com/pi-hole/docker-pi-hole/actions/runs/26350566933/attempts/1?pr=2038), had a 0% cache hit - due to the changes:

image

On the [second run](https://github.com/pi-hole/docker-pi-hole/actions/runs/26350566933?pr=2038), it hit 52%:

image

I was expecting more cache reuse than that. I think there is more opportunities with reorganizing the Dockerfile to make more use build stages - but I don't want to blow up the PR with a bunch of changes that are hard to review.

## Types of changes

## Checklist:

### DIFF
--- a/.github/workflows/build-and-publish.yml
+++ b/.github/workflows/build-and-publish.yml
@@ -61,6 +61,7 @@ jobs:
       setup-qemu: true
       cache: true
       cache-scope: build
+      cache-mode: max
       fail-fast: true
       context: src
       output: image
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -20,7 +20,8 @@ ARG PIHOLE_GID=1000
 ENV DNSMASQ_USER=pihole
 ENV FTL_CMD=no-daemon
 
-RUN apk add --no-cache \
+RUN --mount=type=cache,target=/var/cache/apk \
+    apk add \
     bash \
     bash-completion \
     bind-tools \
@@ -43,21 +44,10 @@ RUN apk add --no-cache \
     unzip \
     wget
 
-# For nightly images, we install gdb and screen for ease of debugging (this is
-# not included in the default image to keep it small), and also prepare the
-# system for a core dump. Furthermore, we already add the required signal
-# instructions to the gdb config file
-RUN if [ "${PIHOLE_DOCKER_TAG}" = "nightly" ]; then \
-    apk add --no-cache gdb screen && \
-    echo "ulimit -c unlimited" >> /etc/profile && \
-    echo "handle SIGHUP nostop SIGPIPE nostop SIGTERM nostop SIG32 nostop SIG33 nostop SIG34 nostop SIG35 nostop SIG36 nostop SIG37 nostop SIG38 nostop SIG39 nostop SIG40 nostop SIG41 nostop" > /root/.gdbinit; \
-    fi
-
-ADD https://ftl.pi-hole.net/macvendor.db /macvendor.db
-COPY crontab.txt /crontab.txt
+ADD --link https://ftl.pi-hole.net/macvendor.db /macvendor.db
 
 # Add PADD to the container, too.
-ADD --chmod=0755 https://raw.githubusercontent.com/${PADD_FORK}/PADD/${PADD_BRANCH}/padd.sh /usr/local/bin/padd
+ADD --link --chmod=0755 https://raw.githubusercontent.com/${PADD_FORK}/PADD/${PADD_BRANCH}/padd.sh /usr/local/bin/padd
 
 # download a the main repos from github
 # if the branch is master we clone the latest tag as sometimes the master branch contains meta changes that have not been tagged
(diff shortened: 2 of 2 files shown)


src/Dockerfile
```diff
@@ -94,11 +84,22 @@ RUN cd /etc/.pihole && \
...
+# For nightly images, we install gdb and screen for ease of debugging (this is
+# not included in the default image to keep it small), and also prepare the
+# system for a core dump. Furthermore, we already add the required signal
+# instructions to the gdb config file
+RUN if [ "${PIHOLE_DOCKER_TAG}" = "nightly" ]; then \
+        apk add --no-cache gdb screen && \
+        echo "ulimit -c unlimited" >> /etc/profile && \
+        echo "handle SIGHUP nostop SIGPIPE nostop SIGTERM nostop SIG32 nostop SIG33 nostop SIG34 nostop SIG35 nostop SIG36 nostop SIG37 nostop SIG38 nostop SIG39 nostop SIG40 nostop SIG41 nostop" > /root/.gdbinit; \
+    fi && \
+    echo "${PIHOLE_DOCKER_TAG}" > /pihole.docker.tag
```
lightswitch05, line 103: Moving this to the end allows more cache reused between nightly and prod builds

.github/workflows/build-and-publish.yml
```diff
@@ -61,6 +61,7 @@ jobs:
       setup-qemu: true
       cache: true
       cache-scope: build
+      cache-mode: max
```
lightswitch05, line 64: Better cache reuse between builds. See [Cache mode](https://docs.docker.com/build/cache/backends/#cache-mode)

src/Dockerfile
```diff
@@ -43,21 +44,10 @@ RUN apk add --no-cache \
...
-    echo "handle SIGHUP nostop SIGPIPE nostop SIGTERM nostop SIG32 nostop SIG33 nostop SIG34 nostop SIG35 nostop SIG36 nostop SIG37 nostop SIG38 nostop SIG39 nostop SIG40 nostop SIG41 nostop" > /root/.gdbinit; \
-    fi
-
-ADD https://ftl.pi-hole.net/macvendor.db /macvendor.db
-COPY crontab.txt /crontab.txt
+ADD --link https://ftl.pi-hole.net/macvendor.db /macvendor.db
 
 # Add PADD to the container, too.
-ADD --chmod=0755 https://raw.githubusercontent.com/${PADD_FORK}/PADD/${PADD_BRANCH}/padd.sh /usr/local/bin/padd
+ADD --link --chmod=0755 https://raw.githubusercontent.com/${PADD_FORK}/PADD/${PADD_BRANCH}/padd.sh /usr/local/bin/padd
```
lightswitch05, line 50: See [--link flag](https://docs.docker.com/reference/dockerfile/#copy---link)

> When --link is used your source files are copied into an empty destination directory. That directory is turned into a layer that is linked on top of your previous state.
>
> Use --link to reuse already built layers in subsequent builds with --cache-from even if the previous layers have changed. This is especially important for multi-stage builds where a COPY --from statement would previously get invalidated if any previous commands in the same stage changed, causing the need to rebuild the intermediate stages again. With --link the layer the previous build generated is reused and merged on top of the new layers. This also means you can easily rebase your images when the base images receive updates, without having to execute the whole build again.

src/Dockerfile
```diff
@@ -94,11 +84,22 @@ RUN cd /etc/.pihole && \
...
+
+# For nightly images, we install gdb and screen for ease of debugging (this is
+# not included in the default image to keep it small), and also prepare the
+# system for a core dump. Furthermore, we already add the required signal
+# instructions to the gdb config file
+RUN if [ "${PIHOLE_DOCKER_TAG}" = "nightly" ]; then \
+        apk add --no-cache gdb screen && \
+        echo "ulimit -c unlimited" >> /etc/profile && \
+        echo "handle SIGHUP nostop SIGPIPE nostop SIGTERM nostop SIG32 nostop SIG33 nostop SIG34 nostop SIG35 nostop SIG36 nostop SIG37 nostop SIG38 nostop SIG39 nostop SIG40 nostop SIG41 nostop" > /root/.gdbinit; \
+    fi && \
```
PromoFaux:
```suggestion
    fi \
```

I think this is probably very low risk/total edgecase - but If we fall into the `if` block, and then something goes wrong inside it (i.e apk add fails for whatever reason) then `/pihole.docker.tag` won't get written.

Does that matter? Is it likely to happen? Doubt it. But maybe it would be clearer to decouple it from the if statement's success (even if it's only an academic change)

lightswitch05: Thank you for reviewing! This is a good point, I'll change it. I think it should be

```suggestion
    fi; \
```

lightswitch05: I've updated the branch

PromoFaux (approved):

PR #2061, opened 2026-07-22
Update BATS core to 1.14

yubiuser:
BATS 1.14. was released today which should bring colors to the CI output

https://github.com/bats-core/bats-core/releases/tag/v1.14.0

### DIFF
--- a/test/run.sh
+++ b/test/run.sh
@@ -58,4 +58,5 @@ if command -v parallel > /dev/null 2>&1; then
     BATS_FLAGS+=("--jobs" "$(nproc)")
 fi
 
-"$BATS" "${BATS_FLAGS[@]}" "${TEST_FILES[@]}"
+# Ensure that the TERM environment variable is set for colored CI output, defaulting to xterm if not set
+TERM=${TERM:-xterm} "$BATS" "${BATS_FLAGS[@]}" "${TEST_FILES[@]}"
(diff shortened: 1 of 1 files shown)


yubiuser:
Before


After

test/run.sh
```diff
@@ -49,15 +49,14 @@ TEST_FILES=(
 # Configure BATS output and parallelization
 BATS_FLAGS=();
 
-# Use pretty output when stdout is a terminal; TAP format for CI
-if [[ -t 1 ]]; then
-    BATS_FLAGS+=("-p")
-fi
+# Use pretty output, since BATS 1.14 CI runs will auto-adjust
+BATS_FLAGS+=("-p")
```
casperklein, line 53: This is not strictly needed, since pretty output is Bats' default. Keeping that flag doesn't hurt, though.

yubiuser: Is it?

https://github.com/bats-core/bats-core/issues/1219

;-)

Added it after I saw your issue, just to be save for then next update. We also use `-p` over in the FTL repo's `test/run.sh`

casperklein: >Added it after I saw your issue,

Ack, I understand. I just wanted to point out, that `--pretty` is normally not needed (unless there is a bug 😆)

rdwebdesign (approved):

PR #2068, opened 2026-08-08
Document the ports for encrypted DNS and HTTP/3

DL6ER:
## Description

Adds the ports that Pi-hole v7 needs when clients use it as an encrypted resolver - DNS-over-TLS on TCP `853` and DNS-over-QUIC on UDP `853` - and the UDP port for HTTP/3 on the web interface. DNS-over-HTTPS gets no port of its own, it is served at `/dns-query` on the HTTPS port, which is noted in the comments.

Also adds `examples/docker-compose-encrypted-dns.yml` and links the new documentation page in the README.

## Motivation and Context

FTL answers DoT/DoH/DoQ itself since pi-hole/FTL#2989 and pi-hole/FTL#3000, with configurable listener ports in pi-hole/FTL#3012, so a container can offer encrypted DNS to the clients in your network without a proxy in front of it - but only if the ports are published. Documentation counterpart: pi-hole/docs#1416.

This is a **draft** to be merged **with the release of v7.0**: the README and the example describe features that have not shipped yet.

## How Has This Been Tested?

Not runtime-tested - this changes comments, an example file and a documentation link only. The example was checked to be valid YAML, and the repository's `editorconfig-checker` and `codespell` both pass.

## Types of changes

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,15 @@ services:
       #- "67:67/udp"
       # Uncomment the line below if you are using Pi-hole as your NTP server
       #- "123:123/udp"
+      # Uncomment the lines below if your clients should use Pi-hole as an encrypted
+      # resolver: DNS-over-TLS (TCP) and DNS-over-QUIC (UDP). DNS-over-HTTPS needs no
+      # port of its own, it is served on the HTTPS port above
+      #- "853:853/tcp"
+      #- "853:853/udp"
+      # Uncomment the line below for HTTP/3, which runs on UDP on the same port
+      # number and carries both the web interface and DoH. Without it, clients
+      # simply stay on HTTP/2
+      #- "443:443/udp"
     environment:
       # Set the appropriate timezone for your location (https://en.wikipedia.org/wiki/List_of_tz_database_time_zones), e.g:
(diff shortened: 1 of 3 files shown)


yubiuser (commented):
I don't think we need an example compose file for encrypted DNS - our default example  in the readme covers it sufficiently. But we can review again when v7 approaches.

rdwebdesign:
Yeah... we don't need both.

I suggest to keep just one of them: or the `docker-compose-encrypted-dns.yml` file, or the comments and ports on the README file.

OneCricketeer:
Hi, just dumb question. So, with this, do we no longer need to deploy Unbound as a sidecar container, for example? Thanks!

darkexplosiveqwx:
> Hi, just dumb question. So, with this, do we no longer need to deploy Unbound as a sidecar container, for example? Thanks! 

Depends. For encrypted DNS (DNS-over-TLS, DNS-over-HTTPS and possibly soon DNS-over-QUIC), no. FTL v7 will be able to act both as a client and server. For recursive DNS resolution you will still require unbound.

PR #2069, opened 2026-08-27
fix: fail closed on FTL and GitHub release downloads

lbellows:
## Summary

`curl` without `--fail` could write an HTTP error body as `pihole-FTL`, or leave `CLONE_BRANCH` as `null` when the GitHub API is rate-limited, producing a broken image.

Require a successful download (`curl -fsSL`) and a real release tag before cloning.

## Test plan

- [x] Extracted `clone_repo` from the Dockerfile and `bash -n` it
- [x] Confirmed FTL download uses `curl -fsSL`
- [x] Confirmed null/empty tag names abort the build
- Full image build / BATS suite was not run here (no Docker daemon)

### DIFF
--- a/src/Dockerfile
+++ b/src/Dockerfile
@@ -59,7 +59,11 @@ RUN clone_repo() { \
         DEST="$4"; \
         CLONE_BRANCH="$BRANCH"; \
         if [ "$BRANCH" = "master" ]; then \
-            CLONE_BRANCH=$(curl -s https://api.github.com/repos/${FORK}/${REPO}/releases/latest | jq -r .tag_name); \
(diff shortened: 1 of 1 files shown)


PromoFaux (approved):

PromoFaux:
Sorry, just noticed that your commit is a different author to your account? Why is that please?

PromoFaux:
It also needs a[ DCO signoff, too,](https://github.com/pi-hole/docker-pi-hole/pull/2069/checks?check_run_id=107546616921) otherwise the merge checks will fail

lbellows:
> Sorry, just noticed that your commit is a different author to your account? Why is that please?

Its another account I have linked through opencode

